PR #33, opened 2022-04-18
Editor work

sunghee-hwang:
I would suggest you to confirm this for members' review at the upcoming call (April 18).

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/pull/33.html" title="Last updated on Apr 18, 2022, 10:54 PM UTC (359295f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/33/eb19960...359295f.html" title="Last updated on Apr 18, 2022, 10:54 PM UTC (359295f)">Diff</a>


cconcolato:
Bikeshed cannot run (see the auto-generated comment above). There are multiple errors:

1) there is a syntax error in the bibliography. There are missing commas after each entry:
```
{
	"RFC8486": {
		"href": "https://datatracker.ietf.org/doc/html/rfc8486",
		"id": "RFC8486",
		"title": "Ambisonics in an Ogg Opus Container",
		"status": "Best Current Practice",
		"publisher": "IETF"
	},
	"RFC7845": {
```
2) the bibliography is missing a reference to AAC and opus_in_isobmff. 
3) I don't know how to fix the use of `[[!CC]]`. 
4) `codec_id` is defined multiple times. same for ambisonics_mode.

tdaede:
Content generally looks good, though I think the syntax tables should be text, not images. My recommendation would be use the same format used for av1-isobmff (which is just code-formatted text): https://github.com/AOMediaCodec/av1-isobmff/blob/main/index.bs#L259

tdaede (approved):
Let's merge this now, and we can then review the open issues to make sure they are closable, and iterate on this base specification.

PR #51, opened 2022-05-02
Fix images in generated document.

tdaede:
Bikeshed only preserves a couple of directories by default.
https://tabatkins.github.io/bikeshed/#echidna-hooks

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/pull/51.html" title="Last updated on May 2, 2022, 10:18 PM UTC (6bc1b7c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/51/b18a7c0...6bc1b7c.html" title="Last updated on May 2, 2022, 10:18 PM UTC (6bc1b7c)">Diff</a>


tdaede:
FWIW the images won't work in the Preview (they don't work on av1-isobmff previews either)

sunghee-hwang (approved):
Looks good to me.

sunghee-hwang:
> FWIW the images won't work in the Preview (they don't work on av1-isobmff previews either)

I think that it is caused by the folder, which index.html is stored in, is different from the folder, which image folder is stored in.
It seems to me that bikeshed tool does not change the file location when it convets index.bs to index.html.

cconcolato:
The problem seems to be the copy to gh-pages.
https://aomediacodec.github.io/iac/images/ does not show anything.

PR #91, opened 2022-08-29
tapas_#1 discussion draft

sunghee-hwang:
All of sections through section 1 ~ 8 are updated as follows:
- To make conformance points be clear
- To resolve comments from IAC SG
- To specify profiles (Simple, Base, Enhanced) and its conformance points
- To make IA bitstream definition be clear
- To make encapsulation and decapsulation scheme be clear
- To make IA encoding and decoding process be clear
- List up some issues and notes.


index.bs
```diff
@@ -631,29 +1107,80 @@ class recon_gain_info_obu() {
...
+class decoder_config(lpcm) {
+  unsigned int (32) sample_rate;
+  unsigned int (8) sample_size;
 }
 ```
 
-<b>Semantics</b>
+<dfn noexport>sample_rate</dfn> shall indicate the sample rate of the input audio in Hz.
+
+<dfn noexport>sample_size</dfn> shall indicate the sample size of the audio sample in bits.
```
tdaede, line 1180: For this to be useful we will need more information (not necessarily in the bitstream), e.g. how the samples are packed and their endianness.

index.bs
```diff
@@ -385,126 +422,381 @@ The reserved OBU allows the extension of this specification with additional OBU
...
+
+<b>Syntax</b>
+
+```
+class global_metadata_obu() {
+  unsigned int (8) version;
+  unsigned int (8) profile_version;
+  unsigned int (8) unique_counter
+  unsigned int (leb128()) num_functions;
+  for (i = 0; i < num_functions; i++) {
```
tdaede, line 565: This seems overly flexible, I think all the contained functions could be in a fixed order.

index.bs
```diff
@@ -385,126 +422,381 @@ The reserved OBU allows the extension of this specification with additional OBU
...
-  unsigned int (2) ambisonics_mode;
-  unsigned int (3) channel_audio_layer;
-  unsigned int (3) reserved;
-  if (ambisonics_mode == 1 or 2)
-    ambix_layer_config(ambisonics_mode);
-  for (i=1; i<= channel_audio_layer; i++)
-    channel_audio_layer_config(i);
+class audio_frame_obu(obu_size) {
+  unsigned int (leb128()) audio_stream_id_ref;
+  unsigned int (leb128()) sync_offset;
```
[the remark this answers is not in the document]
sunghee-hwang: Yes. You can find it right after ISSUE4 in Preview.

index.bs
sunghee-hwang: It is a good point. I will update it after call.
It would be big endian. Don't you think so?

index.bs
podborski: `unsigned int (leb128())` 🤔 This type definition looks a bit strange.

sunghee-hwang: Can I know your kind suggestion for it?

sunghee-hwang: > `unsigned int (leb128())` 🤔 This type definition looks a bit strange.

Can I know your kind suggestion for it?

cconcolato:
The use of "shall" is often not appropriate. For example:
> The descriptor OBUS shall contain all the information that is required to setup and configure the decoders, reconstruction algorithms, renderers and mixers.

It is not a "shall", i.e. not a normative requirement that can be verified by a validation tool. It is a definition, a statement of fact, and should simply use "is" or "are".

"shall" should be used when it is possible to verify if a bitstream verifies this statement (valid bitstream) or does not (invalid bitstream).

Other examples of invalid use of "shall" include:
> obu_type shall specify the type of data structure contained in the OBU payload.

> If obu_type = OBU_IA_Audio_Element, then it shall indicate a unique ID

I suggest doing a pass on the spec and replacing all the inappropriate "shall"s.

index.bs
```diff
@@ -322,189 +295,465 @@ OBU data starts on the first (most significant) bit and ends on the last bit of
...
+ - If obu_type = OBU_IA_Parameter_Block, then it shall indicate a unique ID (i.e. parameter_id) for a given parameter in IA sequence.
+ - If obu_type = OBU_IA_Audio_Frame, then it shall indicate a unique ID (i.e. audio_substream_id) for a given substream in IA sequence. Therefore, all of audio_frame_obus of the same substream shall have the same obu_id.
 
-Reserved units are for future use and shall be ignored by IAC-OBU parser.
 
-<dfn noexport>obu_reserved_1bit</dfn> shall be set to 0. The value is ignored by a IAC-OBU parser.
+<dfn noexport>obu_sync</dfn> shall define the offset from a reference point in the IA bitstream for which the OBU is valid and applicable. The reference point used depends on the IA Profile (See Profiles Section).
 
-<dfn noexport>obu_has_size_field</dfn> equal to 1 indicates that the obu_size syntax element will be present. obu_has_size_field equal to 0 indicates tha the obu_size syntax element will not be present. 
+<dfn noexport>obu_duration</dfn> shall define the duration for which the OBU is valid and applicable.
```
cconcolato, line 351: What is the unit? seconds? milliseconds? something else?

index.bs
```diff
@@ -322,189 +295,465 @@ OBU data starts on the first (most significant) bit and ends on the last bit of
...
 
+<dfn noexport>obu_counter</dfn> shall increment when its payload is different to the previous OBU of the same obu_type. If the payload is identical to the previous OBU of the same obu_type, i.e. it was redundantly copied or repeated in the bitstream, the value of obu_counter shall remain unchanged.
```
cconcolato, line 353: Why is this needed? Error resilience should be provided by the transport layer. Deduplication can be done by comparing bytes.

index.bs
```diff
@@ -322,189 +295,465 @@ OBU data starts on the first (most significant) bit and ends on the last bit of
...
+    }
+  }
+}
+
+class <dfn noexport>codec_config()</dfn> {
   unsigned int (32) codec_id;
   decoder_config(codec_id);
+  unsigned int (leb128()) num_samples_per_frame;
+  unsigned int (leb128()) num_samples_to_trim_at_start;
+  unsigned int (leb128()) num_samples_to_trim_at_end;
```
cconcolato, line 435: This is a fragile design. 
Cutting a bitstream in 2 will make the `trim_at_end` invalid/wrong. Concatenating 2 files will  require editing the second config to remove `trim_at_start`. 
I suggest adding the (optional) information of how many samples are "valid" in an OBU audio frame.

index.bs
```diff
@@ -322,189 +295,465 @@ OBU data starts on the first (most significant) bit and ends on the last bit of
...
+
+For this obu, both of obu_sync_flag and obu_duratrion_flag shall be set to 0.
+
+The metadata in mix_presentation() shall specify how to render and process one or more audio elements. The processed audio elements shall then be summed to generate a single mixed audio signal. Finally, any additional processing specified in the mix_bus_config() metadata shall be applied to the single mixed audio signal in order to generate the final output audio for playback.
+
+
+<b>Syntax</b>
+```
+class mix_presentation_obu() {
+  string mix_presentation_friendly_label;
```
cconcolato, line 531: The language used to express the label should be provided, following IETF BCP 47.

index.bs
```diff
@@ -631,553 +920,524 @@ class recon_gain_info_obu() {
...
+Substream format shall be the audio samples for the frame size.
+
+
+# Profiles # {#profiles}
+
+The IA Profiles define a set of capabilities that are required to decode the coresponding IA sequence.
+
+Additionally, the following restrictions shall apply to all profiles:
+
+0. IA sequence shall composed of a sequence of one or more IA bitstreams.
```
cconcolato, line 1059: The notion of IA Sequence is not defined. Is it really needed? Mix Presentation and IA Bitstream are sufficient I think.

index.bs
```diff
@@ -139,181 +150,143 @@ url: https://en.wikipedia.org/wiki/Q_(number_format); spec: Q-Format; type: dfn;
...
-## OPEN Bitstream Unit (OBU) Syntax and Semantics ## {#obu-section}
+Therefore, the IA bitstream contains information to facilitate the synchronization of the different IA metadata. The synchronizing information in each metadata indicates the offset from a reference point and the duration for which it is valid.
 
-This section specifies the syntax elements and the meaning of the syntax elements read in the <dfn noexport>OBU</dfn> syntax and semantics structures by assuming IAC decoder receives IA bitstream starting with IA Stream Indicator OBU as input. Where, IAC decoder consists of IAC-OBU parer, Substream decoders and Post-Processor.
-- SeenIABitstream initially shall be set to 0.
-- SeenIABitstream shall be set to 1 after receiving IA Stream Indicator OBU.
-- If SeenIABitstream is set to 1, then following OBUs shall be given to the IAC decoder. Otherwise, the OBU will be dropped.
+- <dfn noexport>Sync OBU</dfn> defines reference points in the bitstream, which used to synchronize the data OBUs with each other.
 
-<dfn noexport>drop_obu( )</dfn> is a function call that indicates when IAC decoder should ignore an OBU because it is not assumed as IA bitstream.
```
cconcolato, line 266: The draft uses a mix of `function` (as in the AV1 video spec) and `class` (as used in the ISOBMFF spec). I would suggest sticking to one convention only, adding a "convention" section to document which one was chosen.

index.bs
sunghee-hwang: > Why is this needed? Error resilience should be provided by the transport layer. Deduplication can be done by comparing bytes.

In broadcasting scenario, when global descriptor is placed frequently without being changed, devices can bypass it by using this field.

index.bs
sunghee-hwang: > What is the unit? seconds? milliseconds? something else?

One option would be RFC 5905 NTP Short Format (16 bits(second) + 16 bits(fraction)).

index.bs
sunghee-hwang: > This is a fragile design. Cutting a bitstream in 2 will make the `trim_at_end` invalid/wrong. Concatenating 2 files will require editing the second config to remove `trim_at_start`. I suggest adding the (optional) information of how many samples are "valid" in an OBU audio frame.

It seems to me that the validity of samples can be defined not per audio frame but per a series of audio frames. For example, aac decoders start to output valid samples from 3rd input coded frame.

index.bs
sunghee-hwang: > The notion of IA Sequence is not defined. Is it really needed? Mix Presentation and IA Bitstream are sufficient I think.



> The use of "shall" is often not appropriate. For example:
> 
> > The descriptor OBUS shall contain all the information that is required to setup and configure the decoders, reconstruction algorithms, renderers and mixers.
> 
> It is not a "shall", i.e. not a normative requirement that can be verified by a validation tool. It is a definition, a statement of fact, and should simply use "is" or "are".
> 
> "shall" should be used when it is possible to verify if a bitstream verifies this statement (valid bitstream) or does not (invalid bitstream).
> 
> Other examples of invalid use of "shall" include:
> 
> > obu_type shall specify the type of data structure contained in the OBU payload.
> 
> > If obu_type = OBU_IA_Audio_Element, then it shall indicate a unique ID
> 
> I suggest doing a pass on the spec and replacing all the inappropriate "shall"s.

Thanks for a good suggestion. I will do it.

index.bs
cconcolato: We should use a lossless, integer representation based a timescale value provided in an Description OBU (if these don't need time).

index.bs
cconcolato: In some video codecs, such feature is to the payload (e.g. AVC SPS id) and in AV1 it is not even present (except for Redundant Frame Header). Maybe such optimization is needed. We should be careful about having it at the OBU header level because any OBU, including Audio Frames or Parameters, could start using it. We'd need to have proper restrictions.

index.bs
cconcolato: In your example, you would mark the first 2 samples as having no valid output samples and the 3rd one as having its first X output samples invalid.

index.bs
sunghee-hwang: I have thought that we need to use time values rather than in sample units to cover multiple audio elements with different sampling rates among them.
If we assume they should have the same sampling rate, then we can indicate it in sample units. 
FYI, now, the Description OBU has no timescale value. It was removed.

index.bs
sunghee-hwang: Understood.

index.bs
sunghee-hwang: What I meant is that in that case, all samples of 4th are valid but if we cut it between 3rd and 4th, then all samples of 4th become invalid.

Back to original issue, in cutting case,  we may regard trim_at_end as correct one and in concatenation case, we can leave as it is because two has been encoded separately, there would be no decoding connection between the last frame of one coded stream and the 1st frame of the other. I mean two streams need to be decoded separately.

index.bs
sunghee-hwang: Understood.

index.bs
podborski: The definition of `leb128` from AV1 returns a value and not a number of bits used to parse the value, therefore it looks strange. `leb128` should probably be defined as a new type in IAC spec. Then you could write `leb128 num_functions;` using the SDL syntax. What do you think?

index.bs
sunghee-hwang: Let me correct my self. 
Actually, it was designed when global metadata OBU contains all of codec config, audio element, and mix presentation. But this time as each of them is defined in individual OBU, the counter is nothing valid anymore.
I am realized that It is only valid for redundant parameter block OBU.
Sorry for confusion!

index.bs
sunghee-hwang: Sound great. I will implement it.

index.bs
cconcolato: I'm a bit confused. AFAIK, AAC has 3 complex aspects:
1. Priming: at the start of the stream and only at the start, people typically add extra samples (silence, called priming) before encoding, and then providing trimming information to remove these extra samples after decoding. In ISOBMFF, the amount of samples to remove after decoding is signaled with the edit list
2. Pre-Roll: for similar reasons, when starting to decode for a random sample, the actual decoded signal is only correct after some time. If you actually want to get correct signal from that sample, you have to start decoding from some samples in the past. In ISOBMFF, this is signaled with the `roll` sample group.
3. End trimming. If your input signal does not have a number of audio samples matching a multiple of the the frame length, then you may have extra samples at the end of a frame. In ISOBMFF, this is signaled by providing explicit duration for the last frame and for the track.

For Problem 1 (Priming) and Problem 2 (End Trimming), I suggest indicating at the Audio Frame level if and how many samples need to be discarded and where (beginning or end). This would allow round-tripping: encode a signal to IAC and get back a signal with the same duration and same content (expect for compression artifacts).

For Problem 2 (Pre-Roll), this should be signaled at the Descriptor level, as it is valid for the entire stream and is codec-specific.

sunghee-hwang: Thanks for the suggestion. Now, I see your intention.
I put 2bits flag in obu header for trimming to indicate them at the Audio Frame level.
And, assigned a new field for pre-roll in codec config.
Please check it.

sunghee-hwang:
Let me share the updated list so far since the previous IAC meeting.
- Added convention section
- Start Code OBU: ia_code having 'aiac'
- Codec Config OBU: one per codec config, roll_distance, removed trimming info, list of audio elements which are coded by using the codec config.
- Audio Element OBU: list of parameters which are applied to the Audio Element
- OBU Header: added obu_trimming_flag (2bits)
- Added ID Linking example
- Updated IAC encapsulation & decapsulation schemes.

podborski:
Some comments / questions:

- on syntax: It would be good to align with SDL syntax a bit more. For example use `type memberName;` or `type memberName(constructor values, ...);` instead of just `type();`. Also `memberName` needs to be defined in semantics.
- Missing semantics for flags. Even if its obvious, it is still good to provide semantics for every defined syntax element.
- `obu_trimming_flag` is using 2 bits, is this intended to be 2 bits or is it a typo?
- `f(3)` is not defined in SDL, is it just `bit(3)` or `unsigned int(3)` etc.?

Question, can you explain how the extension mechanism works for `obu_header`, maybe based on an example?

sunghee-hwang:
> Some comments / questions:
> 
> * on syntax: It would be good to align with SDL syntax a bit more. For example use `type memberName;` or `type memberName(constructor values, ...);` instead of just `type();`. Also `memberName` needs to be defined in semantics.
> * Missing semantics for flags. Even if its obvious, it is still good to provide semantics for every defined syntax element.
> * `obu_trimming_flag` is using 2 bits, is this intended to be 2 bits or is it a typo?
> * `f(3)` is not defined in SDL, is it just `bit(3)` or `unsigned int(3)` etc.?
> 
> Question, can you explain how the extension mechanism works for `obu_header`, maybe based on an example?

The mechanism provides the way how devices of the current version are able to skip the extension field of a future version having extended OBU header. 
For example, the future version has obu_extension_flag = 1, then extension_header_size and followed by extension fields for the future syntax. When the future version IA bitstream is input to the devices of the current version, the devices can skip the extension fields from the extension_header_size.

sunghee-hwang:
> Some comments / questions:
> 
> * on syntax: It would be good to align with SDL syntax a bit more. For example use `type memberName;` or `type memberName(constructor values, ...);` instead of just `type();`. Also `memberName` needs to be defined in semantics.
> * Missing semantics for flags. Even if its obvious, it is still good to provide semantics for every defined syntax element.
> * `obu_trimming_flag` is using 2 bits, is this intended to be 2 bits or is it a typo?
> * `f(3)` is not defined in SDL, is it just `bit(3)` or `unsigned int(3)` etc.?
> 

Thank you pointing them out.

Could you help me to update below regarding your first comment.
## Type ## {#type}
The following specifies the type of syntax elements
 <b>leb128()</b>
  Unsigned integer represented by a variable number of little-endian bytes. Its size in bits shall be 8 x the variable number of bytes.
(NOTE: After your updates, I will add leb128() for signed one and others if necessary)

I will update something related to your 2nd and 4th comments.
Regarding the 3rd comment, yes it is intended. I will change the name of syntax as you already made the comment on it.

sunghee-hwang:
Here are the updated list for your reference.

- Convention: BCP 47 needs to be carefully reviewed as I am not 100% sure.
- IAC Decapsulation process moved to the next section of IAC encapsulation
- Added temporal delimiter obu
- audio frame obu
- render_config()
- element_mix_config()

PR #114, opened 2022-10-13
Standalone and obus

felicialim:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/pull/114.html" title="Last updated on Oct 24, 2022, 9:44 PM UTC (9a0dbe2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/114/ec6a802...9a0dbe2.html" title="Last updated on Oct 24, 2022, 9:44 PM UTC (9a0dbe2)">Diff</a>


sunghee-hwang (commented):
- As codec config obu includes audio element ids, it would be better that codec config obus comes before audio element obus.
- There would be no Logistic OBUs. 
- It would be better the timeline for concatenation is based on audio substream only because the timeline for parameter blocks can not be present without audio substreams. So, I vote 2) audio substream only.

felicialim:
I just realized the obu_redundant_copy changes in #119 was supposed to be in here (it appears there as it's based off this branch), so I moved that commit over.

felicialim:
> * As codec config obu includes audio element ids, it would be better that codec config obus comes before audio element obus.
How about switching the reference direction around, so we have the audio element reference the codec config instead (see [2afcf53](https://github.com/AOMediaCodec/iac/pull/114/commits/2afcf53434f0ef192e5adfcb90186f31dfacc199). 

> * There would be no Logistic OBUs.
Done, thanks for catching this. 

> * It would be better the timeline for concatenation is based on audio substream only because the timeline for parameter blocks can not be present without audio substreams. So, I vote 2) audio substream only.
I think we still have to consider the issue of handling overlapping parameters that may happen in that case.

felicialim:
> > * As codec config obu includes audio element ids, it would be better that codec config obus comes before audio element obus.
> 
> How about switching the reference direction around, so we have the audio element reference the codec config instead (see [2afcf53](https://github.com/AOMediaCodec/iac/pull/114/commits/2afcf53434f0ef192e5adfcb90186f31dfacc199).

Reverting for now as it would require changes elsewhere too. Since reordering OBUs is an optimization, I've created a new separate editorial issue (#120), which we can address in a separate PR.  

> 
> > * There would be no Logistic OBUs.
> 
> Done, thanks for catching this.
> 
> > * It would be better the timeline for concatenation is based on audio substream only because the timeline for parameter blocks can not be present without audio substreams. So, I vote 2) audio substream only.
> 
> I think we still have to consider the issue of handling overlapping parameters that may happen in that case.

tdaede (approved):

tdaede:
LMK if there is a preferred order you'd like to land these in.

sunghee-hwang (approved):

PR #118, opened 2022-10-14
Address gaps in the audio signals

felicialim:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/pull/118.html" title="Last updated on Oct 20, 2022, 6:36 PM UTC (03087ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/118/f40f2a5...03087ef.html" title="Last updated on Oct 20, 2022, 6:36 PM UTC (03087ef)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -760,6 +760,8 @@ This value must be greater or equal to 22, in order to avoid collision with the 
 
 <dfn noexport>audio_frame()</dfn> is the raw coded audio data for the frame. It shall be [=opus packet=] of [[!RFC6716]] for IAC-OPUS, [=raw_data_block()=] of [[!AAC]] for IAC-AAC-LC, [=FRAME=] of [[!FLAC]] for IAC-FLAC and audio samples for IAC-LPCM.
 
+For this version of the specification, all audio frames for a given substream must be gapless.
+
 ### Temporal Delimiter OBU Syntax and Semantics ### {#obu-temporaldelimiter}
 
 This section specifies the OBU payload of OBU_IA_Temporal_Delimiter.
@@ -786,7 +788,8 @@ class sync_obu() {
   for (i = 0; i < num_obu_ids; i++) {
     leb128() obu_id;
     unsigned int (1) obu_data_type;
-    unsigned int (7) reserved;
+    unsigned int (1) reinitialize_decoder;
+    unsigned int (6) reserved;
     sleb128() relative_offset;
   }
   leb128() concatenation_rule;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -797,11 +800,27 @@ class sync_obu() {
...
+obu_data_type : Type of OBU data
+      0       : SUBSTREAM
+      1       : PARAMETER
+</pre>
+
+<dfn noexport>reinitialize_decoder</dfn> shall be used to specify the behaviour of a decoder when encountering gaps in the audio substream, where the gap shall be identified as described in [[#standalone-synchronizing-data-obus]]. If obu_data_type does not equal SUBSTREAM, an IAC-OBU parser shall ignore this field.
+
+If reinitialize_decoder = 0, the decoder shall not be reinitialized before decoding the audio frames after the gap. This may be used in the case where it is preferable for the decoder to fill the gap with silence instead.
+
+If reinitialize_decoder = 1, the decoder shall be reinitialized before decoding the audio frames after the gap. If a roll_distance is specified in the relevant Codec Config OBU, it is applicable after reinitializing the decoder.
```
tdaede, line 821: Roll distance is different than preskip - roll is applied during seeking, whereas preskip is applied at the beginning of the stream. I think you want to address preskip here instead.

index.bs
```diff
@@ -797,11 +800,27 @@ class sync_obu() {
...
+
+For this version of the specification, the value of reinitialize_decoder shall be set to 0.
```
tdaede, line 823: We should specify decoder behavior if it sees a 1 here (should it be ignored? stream rejected?)

index.bs
```diff
@@ -2416,3 +2435,27 @@ This section provide a guideline to generate drc_config().
...
+
+In this version of the specification, gaps within an audio substream are not supported.
+
+## A gap between two audio substreams
+
+A gap may occur if there is a period of time between the end of one substream and the start of another. Such a gap may be expressed via the Sync OBU offsets. Similar to the case of a gap within an audio substream, the behaviour of the decoder is determined by the [=reinitialize_decoder=] field in the Sync OBU.
+
+A gap may further occur if there is a period of time between the end of all substreams and the start of another. This case may be expressed by setting a non-zero value for the [=global_offset=] field in the Sync OBU.
+
+## A gap due to packet loss
```
tdaede, line 2459: Just to double check, with this version of the specification, this would always be interpreted to be the case, correct?

index.bs
felicialim: Done.

index.bs
felicialim: Done.

index.bs
felicialim: That's right. I've added a note to spec.

tdaede (approved):

PR #164, opened 2022-11-04
fix #137

### LINKED ISSUE #137: Profiles should be "at least" audiuo elements
Instead of only one and only two audio elements, they should specify at least one or at least two elements.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/pull/164.html" title="Last updated on Nov 11, 2022, 8:05 AM UTC (9d39b25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iac/164/09dbbb4...9d39b25.html" title="Last updated on Nov 11, 2022, 8:05 AM UTC (9d39b25)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1319,7 +1319,7 @@ Restrictions on the IA sequence:
 Capabilities of the IA parser, decoder and processor:
 - They shall be able to parse an IA sequence with the MSB four bits of [=profile_version=] = 0 and the MSB four bits of [=version=] = 0 (i.e., profile_version = 0 to 15 and version = 0 to 15).
 - They shall be able to decode and process up to 16 channels.
-- They shall be able to reconstruct only one audio element.
+- They shall be able to reconstruct at lease one audio element.
 - They may use demixing_info() to do down-mixing.
 
 ## IA Base Profile ## {#profiles-base}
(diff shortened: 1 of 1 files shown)


sunghee-hwang:
> This changes the scope of the profiles as it doesn't place an upper limit on the number of audio elements that can be processed. I'm not sure this is intended?

Requirements of IA sequence are limiting the number of audio elements for profiles.

tdaede:
It does seem that the matching "at most" restriction is missing on the IA Sequence.

I am not opposed to also limiting how many a Base Profile decoder is allowed to decode, however it has some forward compatibility concerns - e.g. does a fancier decoder have to throw away extra audio elements if the stream signals it is Base? Or is the restriction that decoders must support the exact mix of features in one of the Profiles and not more or less?

sunghee-hwang:
> It does seem that the matching "at most" restriction is missing on the IA Sequence.
[SH] I think that the base profile has already it. Please see line 1331. Below is the copy.
```
- There shall be at most two unique Audio Element OBUs at any one time.
```

> I am not opposed to also limiting how many a Base Profile decoder is allowed to decode, however it has some forward compatibility concerns - e.g. does a fancier decoder have to throw away extra audio elements if the stream signals it is Base? Or is the restriction that decoders must support the exact mix of features in one of the Profiles and not more or less?

[SH] My understanding on below sentences is where or not a base profile decoder reconstructs (or mix) more than two audio elements is up to the implementer.
``` 
-They shall be able to reconstruct two audio elements.
- They shall be able to mix two audio elements.
```
If they are read as it is limited that a base profile decoder is allowed to decode only up to two audio elements, I think that it should be revised.

tdaede (approved):
Oh OK, I see that restriction now.

felicialim (approved):
Thanks for making the changes.

Issue #176, opened 2022-11-15
time_base definition is reversed


cconcolato:
The `time_base` definition says:
>time_base shall specify the time base used by this parameter, expressed as seconds per tick. Time-related fields associated with this parameter, such as durations and intervals, shall be expressed in the number of ticks.

I think it specifies the number of ticks per seconds, not the number of seconds per tick.

Issue #175, opened 2022-11-15
param_definition_type should not use leb128()


cconcolato:
It is not a bitstream structure that is read or written. It is a value that is determined at runtime by `get_param_definition_type`. Simply delete `leb128()`.

cconcolato:
Also the table defining `param_definition_type` uses the value 0 three times!

tdaede:
Will be closed once #195 is merged.

PR #195, opened 2022-11-24
Fix #188, #176 and #175

### LINKED ISSUE #188: Merge issue from PR #119 (parameters)
Missing DemixingInfoParamDefinition and ReconGainInfoParamDefinition after merging

sunghee-hwang:
To fix texts related to parameter definitions and types

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/195.html" title="Last updated on Dec 13, 2022, 1:31 AM UTC (2730a76)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/195/9eaebf1...2730a76.html" title="Last updated on Dec 13, 2022, 1:31 AM UTC (2730a76)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -650,26 +650,6 @@ audio_element_type: The type of audio representation.
 
 <dfn noexport>ambisonics_config()</dfn> is a class that provides the metadata required for combining the substreams identified here in order to reconstruct an Ambisonics layout.
 
-#### Parameter Definition Syntax and Semantics #### {#parameter-definition}
-
-Parameter definition classes shall inherit from the abstract <dfn noexport>ParamDefinition()</dfn> class.
-
-<b>Syntax</b>
-
-```
-abstract class ParamDefinition() {
-  leb128() parameter_id;
-  leb128() time_base;
-}
-```
-
-<b>Semantics</b>
-
-<dfn value noexport for="ParamDefinition()">parameter_id</dfn> shall indicate the unique ID in an IA bitstream for a given parameter.
-
-<dfn value noexport for="ParamDefinition()">time_base</dfn> shall specify the time base used by this parameter, expressed as ticks per second. Time-related fields associated with this parameter, such as durations and intervals, shall be expressed in the number of ticks.
-
-
 ### Mix Presentation OBU Syntax and Semantics ### {#obu-mixpresentation}
 
 This section specifies the OBU payload of OBU_IA_Mix_Presentation.
@@ -878,7 +858,6 @@ This value must be greater or equal to 22, in order to avoid collision with the 
 <dfn noexport>audio_frame()</dfn> is the raw coded audio data for the frame. It shall be [=opus packet=] of [[!RFC6716]] for IAC-OPUS, [=raw_data_block()=] of [[!AAC]] for IAC-AAC-LC and [=FRAME=] of [[!FLAC]] for IAC-FLAC.
 
 For IAC-LPCM, [=audio_frame()=] shall be LPCM samples. When more than one byte is used to represent a LPCM sample, the byte order shall be in little endian. 
-
 
 For this version of the specification, all audio frames for a given substream must be gapless.
 
--- a/index.bs
+++ b/index.bs
@@ -571,18 +571,32 @@ class audio_element_obu() {
   
   leb128() num_parameters;
   for (i = 0; i < num_parameters; i++) {
-    leb128() parameter_id;
-    leb128() parameter_name;
+    leb128() param_definition_type;
+    if (param_definition_type == PARAMETER_DEFINITION_DEMIXING_INFO) {
+        DemixingParamDefinition();
+    }
+    if (param_definition_type == PARAMETER_DEFINITION_RECON_GAIN_INFO) {
+        ReconGainParamDefinition();
+    }
   }
 
   if (audio_element_type == CHANNEL_BASED) {
     scalable_channel_layout_config();
   } else if (audio_element_type == SCENE_BASED) {
     ambisonics_config();
-  }
-  
-  
+  }  
 }
+```
+
+```
+class DemixingParamDefintion() extends ParamDefintion() {
+}
+```
+
+```
+class ReconGainParamDefintion() extends ParamDefintion() {
+}
+
 ```
 
 <b>Semantics</b>
@@ -606,18 +620,46 @@ audio_element_type: The type of audio representation.
 
 <dfn value noexport for="audio_element_obu()">parameter_id</dfn> shall be the unique ID associated with a parameter that is used by the algorithm specified in this audio element.
 
-<dfn noexport>parameter_name</dfn> shall specify the name of the parameter.
+<dfn noexport>param_definition_type</dfn> specifies the type of the parameter definition. All parameter definition types described in this version of the specification are listed in the table below, along with their associated parameter definitions.
 
-<pre class = "def">
-parameter_name : Parameter name.
-       0       : SCALABLE_CHANNEL_LAYOUT_DEMIXING_INFO
-       1       : SCALABLE_CHANNEL_LAYOUT_RECON_GAIN_INFO
-   the others  : reserved
-</pre>
+<table class = "def">
+<tr>
+  <th>param_definition_type</th><th>Parameter definition type</th><th>Parameter definition</th>
+</tr>
+<tr>
+  <td>0</td><td>PARAMETER_DEFINITION_MIX_GAIN</td><td>MixGainParamDefinition</td>
+</tr>
+<tr>
+  <td>1</td><td>PARAMETER_DEFINITION_DEMIXING_INFO</td><td>DemixingInfoParamDefinition</td>
+</tr>
+<tr>
+  <td>2</td><td>PARAMETER_DEFINITION_RECON_GAIN_INFO</td><td>ReconGainInfoParamDefinition</td>
+</tr>
+</table>
+
 
 <dfn noexport>scalable_channel_layout_config()</dfn> is a class that provides the metadata required for combining the substreams identified here in order to reconstruct a scalable channel layout.
 
 <dfn noexport>ambisonics_config()</dfn> is a class that provides the metadata required for combining the substreams identified here in order to reconstruct an Ambisonics layout.
+
+#### Parameter Definition Syntax and Semantics #### {#parameter-definition}
+
+Parameter definition classes shall inherit from the abstract <dfn noexport>ParamDefinition()</dfn> class.
+
+<b>Syntax</b>
+
+```
+abstract class ParamDefinition() {
+  leb128() parameter_id;
+  leb128() time_base;
+}
+```
+
+<b>Semantics</b>
+
+<dfn value noexport for="ParamDefinition()">parameter_id</dfn> shall indicate the unique ID in an IA bitstream for a given parameter.
+
+<dfn value noexport for="ParamDefinition()">time_base</dfn> shall specify the time base used by this parameter, expressed as ticks per second. Time-related fields associated with this parameter, such as durations and intervals, shall be expressed in the number of ticks.
 
 
 ### Mix Presentation OBU Syntax and Semantics ### {#obu-mixpresentation}
@@ -700,7 +742,7 @@ class parameter_block_obu() {
   leb128() num_segments;
   leb128() constant_segment_interval;
 
-  leb128() param_definition_type = get_param_definition_type(parameter_id);
+  param_definition_type = get_param_definition_type(parameter_id);
 
   for (i = 0; i < num_segments; i++) {
     if (constant_segment_interval == 0) {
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -571,18 +571,32 @@ class audio_element_obu() {
...
   leb128() num_parameters;
   for (i = 0; i < num_parameters; i++) {
-    leb128() parameter_id;
-    leb128() parameter_name;
+    leb128() param_definition_type;
+    if (param_definition_type == PARAMETER_DEFINITION_DEMIXING_INFO) {
+        DemixingParamDefinition();
+    }
+    if (param_definition_type == PARAMETER_DEFINITION_RECON_GAIN_INFO) {
+        ReconGainParamDefinition();
```
felicialim, line 579: This should be recon_gain_info_parameter_data()

index.bs
```diff
@@ -571,18 +571,32 @@ class audio_element_obu() {
   
   leb128() num_parameters;
   for (i = 0; i < num_parameters; i++) {
-    leb128() parameter_id;
-    leb128() parameter_name;
+    leb128() param_definition_type;
+    if (param_definition_type == PARAMETER_DEFINITION_DEMIXING_INFO) {
+        DemixingParamDefinition();
```
felicialim, line 576: This should be `demixing_info_parameter_data()`, where 

```
class demixing_info_parameter_data() {
  unsigned int (3) dmixp_mode;
  unsigned int (5) reserved;
}
```

index.bs
```diff
@@ -571,18 +571,32 @@ class audio_element_obu() {
   
   leb128() num_parameters;
   for (i = 0; i < num_parameters; i++) {
-    leb128() parameter_id;
-    leb128() parameter_name;
+    leb128() param_definition_type;
```
felicialim, line 574: I think we want to remove this loop and instead, define the parameters closer to where they are actually used, e.g. here we would want to add it in the scalable channel config.

index.bs
```diff
@@ -731,24 +773,9 @@ class parameter_block_obu() {
...
-<tr>
-  <td>0</td><td>PARAMETER_DEFINITION_DEMIXING_INFO</td><td>DemixingInfoParamDefinition</td>
-</tr>
-<tr>
-  <td>0</td><td>PARAMETER_DEFINITION_RECON_GAIN_INFO</td><td>ReconGainInfoParamDefinition</td>
-</tr>
-</table>
-
-NOTE: param_definition_type is not coded in the bitstream but is inferred at run time based on parameter_id.
+Audio Element OBU and/or Mix Presentation OBU shall map a parameter_id to the parameter defintion type. So, IA decoders can know the definition type mapped to the parameter_id.
```
felicialim, line 778: The mapping should be available from ParamDefinition(), which should be inherited by all parameter definition classes. So we shouldn't need an additional mapping here.

felicialim (changes requested):
Thanks for this PR to fix the broken merge. However, it looks like a few things are still missing, I can look into this further.

index.bs
sunghee-hwang: If we move them in the scalable channel config, then when we consider a future extension we may need to define new audio element obu if the future version need a parameter for ambisonics.
For future extendibility, I prefer to keep the current location if there is no big difference.

index.bs
sunghee-hwang: My mistake.
I intended just to explain the mapping information is described in Audio Element OBU and/or Mix Presentation OBU.
I did not intend to define a new mapping.

So, the sentence need to be changed like this:

Audio Element OBU and/or Mix Presentation OBU is mapping a parameter_id to the parameter defintion type. So, IA decoders can know the definition type mapped to the parameter_id.

index.bs
```diff
@@ -731,24 +773,9 @@ class parameter_block_obu() {
...
-<tr>
-  <td>0</td><td>PARAMETER_DEFINITION_DEMIXING_INFO</td><td>DemixingInfoParamDefinition</td>
-</tr>
-<tr>
-  <td>0</td><td>PARAMETER_DEFINITION_RECON_GAIN_INFO</td><td>ReconGainInfoParamDefinition</td>
-</tr>
-</table>
-
-NOTE: param_definition_type is not coded in the bitstream but is inferred at run time based on parameter_id.
+Audio Element OBU and/or Mix Presentation OBU is mapping a parameter_id to the parameter defintion type. So, IA decoders can know the definition type mapped to the parameter_id.
```
felicialim, line 778: Since parameter block parsers have to handle all the different parameter definition types, consider collecting all the mappings between param_definition_type and Parameter definition in a table here instead of distributing them throughout the spec. This would mean there we can look this up in one convenient location in the spec.

index.bs
```diff
@@ -1257,32 +1265,47 @@ class rendering_config() {
 
 ```
 class element_mix_config() {
-  MixGainParamDefinition mix_gain;
+  leb128() mix_gain_type;
```
felicialim, line 1268: The motivation behind allowing the inclusion of the default values in the ParamDefinition() was to provide a clean and consistent way throughout the spec for collecting all the parameter definition related data in one place. It also makes it explicit via syntax that default values can be provided. As such, we would prefer putting the default values back in MixGainParamDefinition().

index.bs
felicialim, line 576: Could you expand on the motivation behind using the DemixingParamDefinition() syntax instead of 
`DemixingParamDefinition demix_param_definition`?

The original motivation for the latter is for reusability and readability. For example, we could have a type of Parameter (e.g. GainParamDefinition) and multiple instances of it for different types of gains. Using the latter syntax, we could then define something like:

```
foo() {
  GainParamDefinition left_channel_gain;
  GainParamDefinition right_channel_gain;
}
```

Of course this generalization wouldn't be required for DemixingParamDefinition(), but we would prefer a single consistent way to express this throughout the spec.

felicialim (changes requested):

index.bs
sunghee-hwang: Optional parameters can be still added without requiring the change of the configuration of an audio element.
Please think it again.

index.bs
sunghee-hwang: Then mix presentation obu is a good candidate to include all of parameter ids to parse

index.bs
sunghee-hwang: Either is ok for me as long as decoders know if the parameter will present in IA sequence from descriptor obus.

index.bs
sunghee-hwang: I think it is a good idea.

index.bs
sunghee-hwang: Copied from the original text.
```
class element_mix_config() {
  MixGainParamDefinition mix_gain;
}

class MixGainParamDefinition extends ParamDefinition() {
  signed int (16) default_mix_gain;
}

default_mix_gain shall specify the default mix gain value to apply when there are no mix gain parameter blocks provided.
```
element_mix_config provides parameter_id which is mapped to mix_gain_parameter_data while MixGainParamDefintion provides default_mix_gain.
Then what does "when there are no mix gain parameter blocks provided." mean?
Does this mean that decoders can not know its presence until getting the parameter_id from IA sequence?
When does decoders need to apply default_mix_gain?
The above sentence creates many following questions. 
Please clarify it.

index.bs
```diff
@@ -606,19 +620,47 @@ audio_element_type: The type of audio representation.
...
-   the others  : reserved
-</pre>
 
 <dfn noexport>scalable_channel_layout_config()</dfn> is a class that provides the metadata required for combining the substreams identified here in order to reconstruct a scalable channel layout.
 
 <dfn noexport>ambisonics_config()</dfn> is a class that provides the metadata required for combining the substreams identified here in order to reconstruct an Ambisonics layout.
 
+#### Parameter Definition Syntax and Semantics #### {#parameter-definition}
+
+Parameter definition classes shall inherit from the abstract <dfn noexport>ParamDefinition()</dfn> class.
```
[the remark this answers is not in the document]
sunghee-hwang: Copied from the original text.

class element_mix_config() {
  MixGainParamDefinition mix_gain;
}

class MixGainParamDefinition extends ParamDefinition() {
  signed int (16) default_mix_gain;
}

default_mix_gain shall specify the default mix gain value to apply when there are no mix gain parameter blocks provided.
element_mix_config provides parameter_id which is mapped to mix_gain_parameter_data while MixGainParamDefintion provides default_mix_gain.
Then what does "when there are no mix gain parameter blocks provided." mean?
Does this mean that decoders can not know its presence until getting the parameter_id from IA sequence?
When does decoders need to apply default_mix_gain?
The above sentence creates many following questions.
it should be clarified.

index.bs
sunghee-hwang: I used demixing_info to align with mix_gain. I also changed demixing_info() class to demixing_info_parameter_data() to align with mix_gain_parameter_data(). Applied the same to recon_gain too.

tdaede:
Felicia is reviewing the updates.

index.bs
```diff
@@ -571,20 +571,34 @@ class audio_element_obu() {
   
   leb128() num_parameters;
   for (i = 0; i < num_parameters; i++) {
-    leb128() parameter_id;
-    leb128() parameter_name;
+    leb128() param_definition_type;
+    if (param_definition_type == PARAMETER_DEFINITION_DEMIXING_INFO) {
+        DemixingParamDefinition demix_info;
```
felicialim, line 576: demix_info -> demixing_info

Or the other way around, to match its definition in the Semantics section

index.bs
```diff
@@ -731,24 +774,9 @@ class parameter_block_obu() {
...
-<tr>
-  <td>0</td><td>PARAMETER_DEFINITION_DEMIXING_INFO</td><td>DemixingInfoParamDefinition</td>
-</tr>
-<tr>
-  <td>0</td><td>PARAMETER_DEFINITION_RECON_GAIN_INFO</td><td>ReconGainInfoParamDefinition</td>
-</tr>
-</table>
-
-NOTE: param_definition_type is not coded in the bitstream but is inferred at run time based on parameter_id.
+Audio Element OBU and/or Mix Presentation OBU is mapping a parameter_id to the parameter definition type. So, IA decoders can know the definition type mapped to the parameter_id.
```
felicialim, line 779: Re: specifying param definitions in the top-level mix presentation OBU, I have a similar comment as specifying them in the top-level audio element OBU. Since there are a few loops in the mix presentation OBU, I'm unsure how easy it is to extend it in the future without changing syntax if we introduced a new param that sits in one of those loops.

index.bs
felicialim: But there are some parameters that apply to audio element, and some to mix presentation?

index.bs
sunghee-hwang: I think that is another story. We should not mix two different issues.

felicialim: Apologies I think I may have misunderstood the original motivation then. My understanding was that it's for future-proofness, was that limited to just ambisonics (and assuming there is no other major changes to ambisonics syntax) and not scalable channel?

sunghee-hwang: I meant that "to define audio element level parameter" differs from "to define layer level parameter". Two are different issues.
"layer level parameter defining scheme" does not cover "audio element level parameter defining scheme" and vise versa.

sunghee-hwang: Currently, as we have no layer level parameters, if someone want to define layer level parameter, then it would be a new one which requires new syntaxes. But, as we already have "audio element level parameter defining scheme", adding a new audio element level parameter can be supported by using the current syntaxes.

index.bs
sunghee-hwang: Yes, right. we need two information as follows:
1. The mapping information between parameter id(s) and an audio element.
2. The mapping information between parameter id(s) and a mixed audio.
 
I think that descriptor obus should provide them without help of data obus so that IA parsers or decoders can make sure all of mapping information before accessing data obus.
For example, when we choose one mix presentation obu, IA parsers or decoders should know all of mapping information without accessing data obus.

index.bs
sunghee-hwang: I think that the current structure would work per audio element id and/or mix presentation id.
```
leb128() num_parameters;
for (i = 0; i < num_parameters; i++) {
    leb128() parameter_id;
    leb128() parameter_name;
}
```
It requires more bits if we have multiple mix presentations while the mix presentation can choose parameters what he want to take.

index.bs
felicialim: I see, this is fine with me. Thanks for clarifying.

index.bs
felicialim: Sorry, I was referring to table in the Semantics section that maps param_definition_type and Parameter definition. It is currently sits in "4.1.7. Audio Element OBU Syntax and Semantics", which seems strange given there are mix parameters there too. But I don't think Mix Presentation OBU semantics is the right place for it either, I think a good place is in the Semantics section of parameter_block_obu(). And perhaps the `param_definition_type` in both audio element OBU and mix presentations OBU can then reference that table. 

As for your comment, I fully agree with it.

index.bs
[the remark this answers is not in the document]
sunghee-hwang: Thanks for the clarification.

As we have the duration field in Parameter Block OBU, it can be covered by below Case 2. It requires more bits but IA parsers or decoders can make sure the situation clearly before accessing data OBUs.
Case1: Default_mix_gain only to cover all of timeline
Case2: Parameter block only to cover all of timeline.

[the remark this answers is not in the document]
felicialim: I think we're missing Case3, where we have parameter blocks covering some of the timeline, and when it doesn't, the default values apply. We should be able to cover Case3 without having to define a new flag for it with the original proposal.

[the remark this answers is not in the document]
felicialim: Since we are defining in element_mix_config(), the IA parsers should be able to know whether there is a default value, before accessing Data OBUs.

index.bs
felicialim: My remaining comment was an editorial one about where the param_definition_type table sits, but I'm happy to create a separate PR to address that to avoid blocking this.

index.bs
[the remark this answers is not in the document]
sunghee-hwang: Do you intend to have all of 3 cases?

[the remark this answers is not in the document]
felicialim: Yes, it would be great if we can offer the flexibility of being able to have a default param value for most of the time, and then every once in a while, a custom value might be applied.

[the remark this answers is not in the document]
sunghee-hwang: If we would have case3, then I don't think that we need case 1&2.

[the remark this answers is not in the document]
sunghee-hwang: I made go back to original case3 only.

felicialim (approved):

tdaede:
Please rebase to make mergeable.

tdaede:
Feel free to merge once rebased.

PR #201, opened 2022-11-29
Fix #198

### LINKED ISSUE #198: coupled_substream_count in channel_audio_layer_config needs to be defined

sunghee-hwang:


index.bs
```diff
@@ -1075,11 +1078,11 @@ If ambisonics_mode is equal to PROJECTION, this shall indicate that the Ambisoni
 
 <dfn noexport>output_channel_count</dfn> shall be the same as [=channel count=] in [[!RFC8486].
 
-<dfn noexport>substream_count</dfn> shall specify the number of audio substreams. It must be the same as [=num_substreams=] in its corresponding audio_element().
+[=substream_count=] shall specify the number of audio substreams. It must be the same as [=num_substreams=] in its corresponding audio_element().
```
felicialim, line 1081: This links to `substream_count` in `scalable_channel_layout_config`, I'm not sure we want that. We have used `dfn value noexport for="...()"` elsewhere, but I'm open to other suggestions too.

index.bs
```diff
@@ -1075,11 +1078,11 @@ If ambisonics_mode is equal to PROJECTION, this shall indicate that the Ambisoni
...
 
 <dfn noexport>channel_mapping</dfn> shall be the same as the one for [=ChannelMappingFamily=] = 2 in [[!RFC8486]].
 
-<dfn noexport>coupled_substream_count</dfn> shall specify the number of referenced substreams that are coded as coupled stereo channels, where M <= N.
+[=coupled_substream_count=] shall specify the number of referenced substreams that are coded as coupled stereo channels, where M <= N.
```
felicialim, line 1085: same as above

index.bs
sunghee-hwang: I tried to do that. But failed due to autolink problem. substream_count is autolinked in many places in the spec. But I don't know how to make it autolink for `dfn value noexport for = "...()"`

tdaede:
These are fixed.

PR #208, opened 2022-12-01
Fix #178

### LINKED ISSUE #178: constant_segment_interval and duration
It's unclear if the sum of all `segment_interval` values (when `constant_segment_interval` is not set) shall equal or not the `duration` value in the parameter_block_obu. If this is the case, the spec should say it. If not, the spec should indicate the behavior in case of mismatch.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/208.html" title="Last updated on Dec 1, 2022, 6:25 AM UTC (4bcb609)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/208/678eed5...4bcb609.html" title="Last updated on Dec 1, 2022, 6:25 AM UTC (4bcb609)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -735,11 +735,11 @@ class parameter_block_obu() {
 
 <dfn value noexport for="parameter_block_obu()">parameter_id</dfn> shall indicate the unique ID that is associated with a specific parameter definition. All parameter blocks that provide data for that parameter definition shall have the same parameter_id.
 
-<dfn noexport>duration</dfn> shall specify the duration for which this parameter block is valid and applicable. The duration shall be expressed as the number of ticks at the rate indicated by the time_base specified in the corresponding parameter definition.
+<dfn noexport>duration</dfn> shall specify the duration for which this parameter block is valid and applicable. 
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -762,6 +762,12 @@ NOTE: param_definition_type is not coded in the bitstream but is inferred at run
 
 <dfn noexport>segment_interval</dfn> shall specify the interval for the given segment.
 
+Each value of [=duration=], [=constant_segment_interval=] and [=segment_interval=] shall be expressed as the number of ticks at the rate indicated by the time base specified in the corresponding parameter definition.
+- When it defines <dfn noexport>D</dfn> = the value of [=duration=], <dfn noexport>NS</dfn> = the value of [=num_segments=], <dfn noexport>CSI</dfn> = the value of [=constant_segment_interval=] and <dfn noexport>SI</dfn> = the value of [=segment_interval=].
```
felicialim, line 766: I suggest replacing "When it defines" with "Let"

felicialim (approved):

PR #221, opened 2022-12-15
Fix #219

### LINKED ISSUE #219: Clarify the location of descriptor OBUs in terms of short-lived contents.
For example, 
- 1st timeline: Audio Element A only
- 2nd timeline: Audio Element A + Audio Element B (commentary)
- 3rd timeline: Audio Element A
In this case, where should descriptor OBUs be present? How does Mix Presentation looks like for each timeline?

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/221.html" title="Last updated on Dec 15, 2022, 3:17 AM UTC (ddd6c60)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/221/77f7ada...ddd6c60.html" title="Last updated on Dec 15, 2022, 3:17 AM UTC (ddd6c60)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1567,6 +1567,8 @@ Additionally, the following constraints apply to the Audio Frame and Parameter B
 
 The above describes the full sequence of OBUs for a given set of descriptor OBUs and their associated data OBUs. If the IAC configuration changes, a new set of descriptor OBUs is required. In that case, a new sequence of the complete set of descriptor OBUs, a Sync OBU and their corresponding data OBUs shall follow, in the same order as described above.
 
+NOTE: In case of that IA sequence contains two audio elements which one of them is a short-lived contents, the number of audio elements is changed (i.e. one to two or two to one) depending on whether the short-lived contents is present or not. So, a new set of descriptor OBUs presents to indicate the changes. In this case, the new set of descriptor OBUs includes a new Mix Presentation OBU to provide the proper mixing according to the audio element(s) to be mixed.
+
 The descriptor OBUs may additionally be repeated redundantly and as frequently as necessary. In this case, the "obu_redundant_copy" field in the OBU header of each of the descriptor OBUs shall be set to 1.
 
 If there is set of descriptor OBUs placed mid-stream, there may be parameter blocks that came before them which are still valid and applicable for the duration after the descriptor OBUs. In this case, these parameter blocks must be redundantly copied and placed after the first Sync OBU that follows the descriptor OBUs. This ensures that any receiver joining mid-stream and encountering a set of descriptor OBU is guaranteed to be able to receive the complete set of metadata that is applicable to all audio frames that come after.


index.bs
```diff
@@ -1567,6 +1567,8 @@ Additionally, the following constraints apply to the Audio Frame and Parameter B
 
 The above describes the full sequence of OBUs for a given set of descriptor OBUs and their associated data OBUs. If the IAC configuration changes, a new set of descriptor OBUs is required. In that case, a new sequence of the complete set of descriptor OBUs, a Sync OBU and their corresponding data OBUs shall follow, in the same order as described above.
 
+NOTE: In case of that IA sequence contains two audio elements which one of them is a short-lived contents, the number of audio elements is changed (i.e. one to two or two to one) depending on whether the short-lived contents is present or not. So, a new set of descriptor OBUs presents to indicate the changes. In this case, the new set of descriptor OBUs includes a new Mix Presentation OBU to provide the proper mixing according to the audio element(s) to be mixed.
```
felicialim, line 1570: We should also make it clear that this is not the only way to signal short-lived contents. The other way uses Sync OBUs, without requiring a new Mix Presentation. This is done via placing a new Sync OBU when the number of audio elements changes, where the list of obu_ids in the Sync OBU changes to reflect which audio elements are present in the timeline that comes immediately afterwards.

sunghee-hwang: Maybe, the word "new" in new mix presentation OBU seems create a confusion. I would replace it to "updated".

Whenever a new set of descriptor OBUs presents, a new Sync OBU shall follow regardless of the new set of descriptor OBUs includes an updated mix presentation.
I am not sure what is the case. But there may or may not be a case which does not require a new set of descriptor OBUs and an updated mix presentation so that only a new sync obu can indicate the situation properly.

Does the changes of audio elements require the change of the number of audio elements to be mixed (i.e the update of mix presentation)?

sunghee-hwang: I understand the mechanism of Sync OBU which can indicate the status of audio elements which will come.
What I am interesting here is decoder operation based on the mix presenation.

If we do not provide a new mix presentation, in case of 2, the mix presentation is defining element_mix for each audio element, output_mix for mixed audio and loudness information of the mixed audio on the measure layout.
Then, how should decoders use the mix presentation on the timeline when only the first long-lived audio element presents? (i.e. for the case of that only the first long-lived audio element is playbacked.)

felicialim: If we have one mix presentation only and only the first long-lived audio element is present, then:

- Output mix is still applied similarly, except that now we only have one long-lived audio element to contribute to the summation when creating the final mix for this section.

- The same loudness info still applies to it similarly as when we playback the section with both audio elements present. This is because the loudness is measured on the final (sub)mix of whole section of the audio specified by that mix presentation. In this case, we're just splitting up the final audio mix and describing the individual mix components separately in the spec.

Btw, just wanted to clarify that your original change to add a "NOTE" about updating mix presentations looks good to me. I'm just suggesting that if we include a note about how to update for short-lived audio elements, maybe it's also worth adding that mix presentations isn't the only way to do it. If you agree, we could also create a new issue and add it in a separate PR if you prefer.

sunghee-hwang: It's fine for me.

tdaede (approved):

Issue #226, opened 2023-01-12
Clarify how short-lived audio elements are signalled 


felicialim:
There are two ways to signal short-lived audio elements - through mix presentations and sync OBUs. Expand the current note in the draft to include both methods.

sunghee-hwang:
It is closed as issue #227 is already resolved.

PR #227, opened 2023-01-12
Clarify how short-lived audio elements are signalled.

felicialim:
Addresses #226

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/227.html" title="Last updated on Jan 24, 2023, 10:40 PM UTC (eae1d21)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/227/d7bdaa1...eae1d21.html" title="Last updated on Jan 24, 2023, 10:40 PM UTC (eae1d21)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2916,3 +2916,5 @@ An IA sequence may contain audio elements which are short-lived, where the audio
 1. The first method uses a new set of Descriptor OBUs. The first set of Descriptor OBUs includes an Audio Element OBU for each of A and B, and a Mix Presentation OBU that references both A and B. When the audio element B has ended, a second set of Descriptor OBUs is placed in the IA stream, which includes an Audio Element for only A, and a Mix Presentation OBU that references only A.
 
 2. The second method uses Sync OBUs, without requiring a new Mix Presentation OBU. The Mix Presentation OBU contains an Audio Element OBU for each of A and B, and a Mix Presentation OBU that references both A and B. The first Sync OBU placed immediately after the Mix Presentation OBU includes both A and B. When the audio element B has ended, a new Sync OBU is placed in the IA stream which includes only A, indicating that B is no longer present in the timeline that comes immediately afterwards.
+
+In both cases, relevant parameter blocks must be updated in a way to match the timing of the short-lived audio elements if necessary to ensure that the processed audio output is correct.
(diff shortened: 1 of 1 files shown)


felicialim:
I included an example but please let me know if you prefer to keep the description more general instead.

index.bs
```diff
@@ -2910,3 +2908,11 @@ A gap may further occur if there is a period of time between the end of all subs
...
+
+1. The first method uses a new set of Descriptor OBUs. The first set of Descriptor OBUs includes an Audio Element OBU for each of A and B, and a Mix Presentation OBU that references both A and B. When the audio element B has ended, a second set of Descriptor OBUs is placed in the IA stream, which includes an Audio Element for only A, and a Mix Presentation OBU that references only A.
+
+2. The second method uses Sync OBUs, without requiring a new Mix Presentation OBU. The Mix Presentation OBU contains an Audio Element OBU for each of A and B, and a Mix Presentation OBU that references both A and B. The first Sync OBU placed immediately after the Mix Presentation OBU includes both A and B. When the audio element B has ended, a new Sync OBU is placed in the IA stream which includes only A, indicating that B is no longer present in the timeline that comes immediately afterwards.
```
sunghee-hwang, line 2918: In the method 2, how to render and mix A with the Mix Presentation OBU (includes both A and B) during the timeline that B is no longer present?
During the timeline, the Mix Presentation OBU is still requiring element_mix config and output_mix_config of A even though decoders ignore the processing of B. But the element_mix_config and output_mix_config were designed for the mixed signal of A and B.

And also whatever it is or a third one, it would be preferred one unified rule rather than 2 alternative ones.

felicialim: It would simply ignore the processing of B since there is no audio in B to apply it to. 

- `element_mix_config` for B is not used when B is no longer present
-  `output_mix_config` applies to the final mix, but in this case the final mix only consists of A.

sunghee-hwang: What I was trying to say was that. 1) When A is only playback, I assume that the gain value of element_mix_config for example will be 0. But 2) in the above case (method 2), the gain value will be set by assuming the mixing with B so that it will not be 0. So, the output signal of 2) will not be big enough relatively compared to the one of 1).
I am not sure if it is a proper processing.
What do you think?

tdaede: We'll require the encoder to update gain with parameter blocks timed in a way to match the short-lived audio element.

tdaede (approved):
Approved once text is updated.

sunghee-hwang (approved):

PR #241, opened 2023-01-20
Fix #170

### LINKED ISSUE #170: Better use of "shall"
The use of "shall" should be reserved to cases where the subject is a bitstream or a construct and the corresponding statement can be verified/validated. In other words, if there is a "shall" it should be possible to create a bitstream/file that is valid against the shall and another bitstream/file that is invalid against the "shall". There could also be cases where "shall" applies to a reader.

The current draft uses "shall" for rules that cannot be verified because they are definitions or statement of facts on the spec.
Examples:
> All of syntax elements shall conform to SDL
This should simply be:
> All of syntax elements conform to SDL

Another example:
> OBU data shall start on the first (most significant) bit and shall end on the last bit of the given bytes. The payload of an OBU shall lie between the first bit of the given bytes and the last bit before the first zero bit of the byte_alignment().
This is by definition. The payload of an OBU is the data between the first and last byte excluding the header. I would remove shall or even remove the entire paragraph.

> obu_size shall indicate the size in bytes
Again a statement of fact. Use "indicates".

There are many other occurrences. A pass on the spec should be done.

sunghee-hwang:
For the better use of "shall"

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/241.html" title="Last updated on Feb 7, 2023, 2:39 AM UTC (a6251f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/241/757baa8...a6251f3.html" title="Last updated on Feb 7, 2023, 2:39 AM UTC (a6251f3)">Diff</a>


tdaede:
Needs rebase and review.

index.bs
```diff
@@ -214,29 +214,29 @@ url: https://xiph.org/flac/format.html; spec: FLAC; type: property;
...
 
- <b>sleb128()</b> indicates the type of an signed integer. It indicates the following signed integer <b>syntaxName</b> shall be encoded by [=leb128()=] specified in [[!AV1-Convention]].
+ <b>sleb128()</b> indicates the type of an signed integer. It indicates the following signed integer <b>syntaxName</b> is encoded by [=leb128()=] specified in [[!AV1-Convention]].
  
  <b>syntaxName</b> is an signed integer which is encoded by [=leb128()=] specified in [[!AV1-Convention]].
  
  <b>string</b> <b>syntaxName</b>
 
- <b>string</b> indicates the type of a string with ring which is terminated by null of one byte (i.e. 0x00).
+ <b>string</b> indicates the type of a string with string which is terminated by null of one byte (i.e. 0x00).
```
felicialim, line 235: Should this just be "type of a string which is"

index.bs
```diff
@@ -484,21 +482,21 @@ It shall always be set to 0 for the following obu_type values:
 - OBU_IA_Audio_Frame
 - OBU_IA_Audio_Frame_ID0 to OBU_IA_Audio_Frame_ID21
 
-<dfn noexport>obu_trimming_status_flag</dfn> indicates whether this OBU has audio samples to be trimmed or not. If it is set to 1, the [=num_samples_to_trim_at_start=] and [=num_samples_to_trim_at_end=] fields shall be present.
```
felicialim, line 487: Perhaps this is a valid use of "shall be", since we can create valid and invalid bitstreams against this case.

index.bs
```diff
@@ -484,21 +482,21 @@ It shall always be set to 0 for the following obu_type values:
...
+<dfn noexport>obu_trimming_status_flag</dfn> indicates whether this OBU has audio samples to be trimmed or not. If it is set to 1, the [=num_samples_to_trim_at_start=] and [=num_samples_to_trim_at_end=] fields present.
 
-<dfn noexport>obu_extension_flag</dfn> indicates whether the [=extension_header_size=] field shall be present. If it set to 0, the [=extension_header_size=] field shall not be present. Otherwise, the [=extension_header_size=] field shall be present.
```
felicialim, line 489: Same here.

index.bs
```diff
@@ -2524,7 +2522,7 @@ Input audio shall be one of followings:
 
 User inputs are:
 - Ambisonics mode to indicate if [=ChannelMappingFamily=] = 2 or 3 of [[RFC8486]].
-- List of channel layouts to be supported for scalable channel audio: it shall conform to [=loudspeaker_layout=].
+- List of channel layouts to be supported for scalable channel audio: each of channel layouts shall conform to [=loudspeaker_layout=].
```
felicialim, line 2525: perhaps "shall conform" could be changed here to "conforms" too?

felicialim, line 2525: perhaps "shall conform" could be changed here to "conforms" too?

index.bs
sunghee-hwang: I am confused which part you want to update

index.bs
sunghee-hwang: I am confused which part you want to update

index.bs
sunghee-hwang: sorry for being late to get your point here. It is updated.

index.bs
sunghee-hwang: sorry for being late to get your point here. It is updated.

felicialim (approved):

tdaede:
OK to merge after rebase

PR #244, opened 2023-01-24
Clarify how the Bezier params should be applied to audio samples.

### LINKED ISSUE #223: Clarify quadratic interpolation equation in animated parameters
See discussion on: https://github.com/AOMediaCodec/iamf/pull/222#discussion_r1057020711

felicialim:
Fixes #223

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/244.html" title="Last updated on Jan 30, 2023, 10:44 PM UTC (d84f825)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/244/13ec14c...d84f825.html" title="Last updated on Jan 30, 2023, 10:44 PM UTC (d84f825)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -897,7 +897,7 @@ Audio Element OBU and/or Mix Presentation OBU is mapping a parameter_id to the p
 
 <dfn noexport>segment_interval</dfn> shall specify the interval for the given segment.
 
-Each value of [=duration=], [=constant_segment_interval=] and [=segment_interval=] shall be expressed as the number of ticks at the rate indicated by the time base specified in the corresponding parameter definition.
+Each value of [=duration=], [=constant_segment_interval=] and [=segment_interval=] shall be expressed as the number of ticks at the [=parameter_rate=] specified in the corresponding parameter definition.
 - When it defines <dfn noexport>D</dfn> = the value of [=duration=], <dfn noexport>NS</dfn> = the value of [=num_segments=], <dfn noexport>CSI</dfn> = the value of [=constant_segment_interval=] and <dfn noexport>SI</dfn> = the value of [=segment_interval=].
 	- When [=CSI=] != 0, [=NS=] x [=CSI=] shall be equal to or greater than [=D=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2445,36 +2445,52 @@ The rendered and processed audio elements are then summed, and then apply [=outp
...
+P2 = (t2, end_point_value),
+```
+
+where <code>t0 = 0</code> is the segment start time, <code>t2 = T - 1</code> and <code>t1</code> is the control point time given by
 
 ```
-P0 = (t_start, start_point_value),
-P1 = (t_control, control_point_value),
-P2 = (t_end, end_point_value),
+t1 = round(T * control_point_relative_time) - 1.
```
tdaede, line 2463: Would be good to define round() exactly (e.g. behavior for 0.5 and -0.5)

tdaede:
We'll review the bezier function definitions a bit more before merging.

tdaede:
OK, so I looked at this in a bit more detail. Below I will describe x as time and y as parameter value.

So `a` in the original text is the Bezier interval. It corresponds to neither x nor y but has to be interpolated to get both.

The new method gives values for `a` but the problem is that they do not map to values of x directly. There are methods to plot / approximate a curve. This PR ends up defining one but because a is the bezier interval, not x, it will not create values for all values of x and will need some sort of interpolation.

In general, you cannot take a bezier curve and make it in the form y = f(x) because that is not guaranteed to be a function, a bezier curve may have multiple values for one x (e.g. it overlaps itself). With the limited way that we define the interpolation, I believe this is not possible. Therefore, one should be able to solve the quadratic to get it in this form, e.g. determine a based on y, then x based on a.

sunghee-hwang (approved):
'n's on 'alpha' and 'beta' are erased eventually.
So, they should be
alpha = n0 - 2 * n1 + n2
beta = 2 * (n1 - n0)

tdaede (approved):

PR #300, opened 2023-03-01
Fix #288-Managing of multiple trimming information

### LINKED ISSUE #288: Need to clarify how to handle multiple trimming informations
Trimming informations can be set to multiple substreams and its relevant ISOBMFF boxes.
We need to say all trimming information shall be same for a given contents.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/300.html" title="Last updated on Mar 7, 2023, 2:56 AM UTC (7dbeaf0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/300/b4e2394...7dbeaf0.html" title="Last updated on Mar 7, 2023, 2:56 AM UTC (7dbeaf0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -85,6 +85,7 @@ url: https://tools.ietf.org/html/rfc8486#; spec: RFC8486; type: dfn;
 url: https://tools.ietf.org/html/rfc7845#; spec: RFC7845; type: dfn;
 	text: ID Header
 	text: Output Gain
+	text: Pre-skip
 
 url: https://tools.ietf.org/html/rfc6716#; spec: RFC6716; type: dfn;
 	text: opus packet
@@ -1529,6 +1530,7 @@ For legacy codecs, Decoder_Config() shall be exactly the same information as the
 
 Codec_Specific_Info for IAMF-OPUS conforms to [=ID Header=] with [=ChannelMappingFamily=] = 0 of [[!RFC7845]] with following constraints:
 - [=Channel Count=] must be set to 2. [=Channel Count=] can be ignored because the real value can be determined from the Audio Element OBU and from the opus packet header.
+- [=Pre-skip=] shall be same as the number of audio samples to be trimmed at the start of substreams.
 - [=Output Gain=] shall not be used. In other words, it shall be set to 0dB.
 - The byte order of each field in [=ID Header=] is converted to big endian.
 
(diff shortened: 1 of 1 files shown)


tdaede:
I think we need a bot more specific language to define substreams which start at the beginning and end of the file. In particular, the end time of the file is itself defined by the substreams, so it's a bit circular.

tdaede:
For defining substreams that start and end at the beginning or end of the file, is this defined in terms of outputting the first and last audio sample of the stream, or containing audio packets at the start and end of the stream?

I would suggest the latter definition.

If ISOBMFF is limited to the base profile where trimming information is the same, this is less ambiguous.

index.bs
```diff
@@ -1830,8 +1834,9 @@ This section describes the basic data structures used to signal encapsulation of
 
 ### Requirement of IA sequence ### {#isobmff-singletrack-iasequence}
 
-IA sequence shall comply with the bitstream which is specified in [[#profiles-simple]] or [[#profiles-base]] for encapsulation of ISOBMFF with single track.
-
+IA sequence shall comply with the bitstream which is specified in [[#profiles-simple]] or [[#profiles-base]] with following restrictions for encapsulation of ISOBMFF with single track.
+- All of substreams which start at the start of the file shall have the same number of audio samples to be trimmed.
+- All of substreams which end at the end of the file shall have the same number of audio samples to be trimmed.
```
shawnisingh, line 1839: for discussion - should we consider having these constraints apply to the profiles?   I feel it may be a good idea for us to avoid unnecessary constraints in ISOBMFF integration, instead preferring to have those constraints defined in the profile.

sunghee-hwang: I think that Shawn's comments simplify it.
Also, for short-lived contents it would not be a big deal even if we restrict its trimming information to be same as that of audio packets at the start and end of the stream, respectively. Actually, the same trimming info at start of short-lived contents as that at the start of the stream would provide more beneficial such as easy frame by frame synchronization rather than restriction.  
 So, I updated it as follows;
1. All substreams shall have the same trimming information for both simple and base profiles
2. Removed restrictions in ISOBMFF.

tdaede (approved):

PR #301, opened 2023-03-02
Fix #291, Updated parameter definition

### LINKED ISSUE #291: Consider updating ParameterDefinition to support the case of that all Parameber block OBUs have the same configuration information.
If we put a flag in Parameter Definition to indicate  that all parameter block OBUs have the same duration, num_segments, constant_segment_interval and segment_interval, every Parameter Block OBU does not need to have the four fields.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/301.html" title="Last updated on Mar 14, 2023, 12:36 AM UTC (c709a2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/301/908a94f...c709a2e.html" title="Last updated on Mar 14, 2023, 12:36 AM UTC (c709a2e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -899,15 +899,20 @@ The metadata specified in this OBU defines the parameter values for an algorithm
 ```
 class parameter_block_obu() {
   leb128() parameter_id;
-  leb128() duration;
-  leb128() num_segments;
-  leb128() constant_segment_interval;
-
-  param_definition_type = get_param_definition_type(parameter_id);
+  
+  (param_definition_type, param_definition_mode, duration, num_segments, constant_segment_interval, segment_interval) = get_param_definition(parameter_id);
+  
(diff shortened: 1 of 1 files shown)


tdaede:
@jwcullen and I will analyze this to make sure it is compatible with random access.

sunghee-hwang:
> @jwcullen and I will analyze this to make sure it is compatible with random access.

As Shawn mentioned during the call, as parameter definitions are stored in Descriptor OBUs and random access needs to start with descriptor OBUs, I think that it gives no impact to random access.

tdaede (approved):
OK to merge when @yeroro's comment addressed.

PR #316, opened 2023-03-09
Fix #297, Rigidly specified order for Descriptors

### LINKED ISSUE #297: Rigidly specify descriptor OBU order regardless of where they appear in the bitstream
The order of the descriptor OBUs is specified in section 6.1.1 for the beginning of the bitstream, but there are cases where they can appear in other parts of the bitstream (6.1.3). We should specify the ordering consistently for all of these cases.

We should also determine if we should repeat all of the descriptor OBUs at once.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/316.html" title="Last updated on Mar 9, 2023, 7:15 AM UTC (c5760d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/316/4621c48...c5760d0.html" title="Last updated on Mar 9, 2023, 7:15 AM UTC (c5760d0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1705,7 +1705,7 @@ This section details the order in which the OBUs are sequenced in a standalone I
 An IA sequence is composed of a series of OBUs in the sequence of a set of descriptor OBUs followed by their associated data OBUs, and where this pattern is repeated as many times as needed.
 
 ### Descriptor OBUs ### {#standalone-descriptor-obus}
-A set of Descriptor OBUs shall be placed at the beginning of the bitstream in the following order:
+A set of Descriptor OBUs shall be placed in the following order regardless of where they appear in the bitstream:
 
 1. One Magic Code OBU
 2. All Codec Config OBUs


index.bs
```diff
@@ -1705,7 +1705,7 @@ This section details the order in which the OBUs are sequenced in a standalone I
 An IA sequence is composed of a series of OBUs in the sequence of a set of descriptor OBUs followed by their associated data OBUs, and where this pattern is repeated as many times as needed.
 
 ### Descriptor OBUs ### {#standalone-descriptor-obus}
-A set of Descriptor OBUs shall be placed at the beginning of the bitstream in the following order:
+A set of Descriptor OBUs shall be placed in the following order regardless of where they appear in the bitstream:
 
 1. One Magic Code OBU
 2. All Codec Config OBUs
```
yeroro, line 1711: So in the case of base profile, which only allows one unique Codec Config OBU but many other descriptor OBUs (refreshing is allowed), should the Codec Config OBU be repeated redundantly? Or can it be missing from the refreshing set?

yeroro (changes requested):

tdaede:
Please address yeroro's comment, but the change looks good otherwise.

index.bs
sunghee-hwang: The Codec Config OBU should be repeated redundantly

yeroro (approved):

PR #325, opened 2023-03-14
Fix #323, fixed typo on loudness_info().

### LINKED ISSUE #323: Fix the bug in loudness_info()
when info_type = 0, the true peak is defined. But the implemented code is wrong.
if (info_type & 1) {
    signed int (16) true_peak;
}
It should be changed to
if (info_type == 0) {
    signed int (16) true_peak;
}

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/325.html" title="Last updated on Mar 14, 2023, 11:23 PM UTC (e947ae7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/325/d055820...e947ae7.html" title="Last updated on Mar 14, 2023, 11:23 PM UTC (e947ae7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -865,7 +865,7 @@ class loudness_info() {
   signed int (16) integrated_loudness;
   signed int (16) digital_peak;
 
-  if (info_type & 1) {
+  if (info_type == 0) {
     signed int (16) true_peak;
   }
 }


yeroro (approved):

jwcullen (changes requested):
See my comment on the issue (#323). 

1) I think it was correct without the change and issue #323 should be closed and this PR should be rejected.

2) It may be better for backwards compatibility to keep this as a bitmask which is how it is currently worded.

3) If this PR merges in I think the semantics wording for `info_type` below are incorrect. As the table refers to bits being set and not checking for specific values.

sunghee-hwang:
You are right. 
Let me close this PR as I misunderstood it.

PR #327, opened 2023-03-15
Fix #326, semantic of obu_size

### LINKED ISSUE #326: Need to reconsider or clarify the semantics of obu_size field
The obu_size in the current WD includes the bytes within the obu_size syntax field.
So, in some cases, we need to recalculate the final value of "leb128() obu_size".
NOTE: The obu_size in AV1 is not including the bytes within the obu_size syntax field.

[As Is]
```
obu_size indicates the size in bytes of the OBU not including the bytes within the obu_header of the preceding fields, i.e. obu_type, obu_redundant_copy, obu_trimming_status_flag and obu_extension_flag.
```
[To Be]
```
obu_size indicates the size in bytes immediately following the OBU's obu_size field.
```
And also, we need to delete the below sentence in Magic Code OBU.
```
For this obu, the obu header (2 bytes) shall be set to 0xF807 (in case of obu_redundant_copy = 0) or FC07 (in case of obu_redundant_copy = 1) for this version of the specification.
```

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/327.html" title="Last updated on Mar 15, 2023, 4:01 PM UTC (bfb07ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/327/d055820...bfb07ae.html" title="Last updated on Mar 15, 2023, 4:01 PM UTC (bfb07ae)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -540,7 +540,7 @@ This flag shall be set to 0 for the current version of the specification (i.e. [
 
 NOTE: A future version of specification may use this flag to specify an extension header field by setting [=obu_extension_flag=] = 1 and setting the size of extended header to [=extension_header_size=].
 
-<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU immediately following the obu _size field of the OBU.
+<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU immediately following the obu_size field of the OBU.
 	
 <dfn noexport>num_samples_to_trim_at_start</dfn> indicates the number of samples that needs to be trimmed from the start of the samples in this Audio Frame OBU. 
 
--- a/index.bs
+++ b/index.bs
@@ -540,8 +540,8 @@ This flag shall be set to 0 for the current version of the specification (i.e. [
 
 NOTE: A future version of specification may use this flag to specify an extension header field by setting [=obu_extension_flag=] = 1 and setting the size of extended header to [=extension_header_size=].
 
-<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU not including the bytes within the obu_header of the preceding fields, i.e. obu_type, obu_redundant_copy, obu_trimming_status_flag and obu_extension_flag.
-
+<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU immediately following the obu _size field of the OBU.
+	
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -540,8 +540,8 @@ This flag shall be set to 0 for the current version of the specification (i.e. [
 
 NOTE: A future version of specification may use this flag to specify an extension header field by setting [=obu_extension_flag=] = 1 and setting the size of extended header to [=extension_header_size=].
 
-<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU not including the bytes within the obu_header of the preceding fields, i.e. obu_type, obu_redundant_copy, obu_trimming_status_flag and obu_extension_flag.
-
+<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU immediately following the obu _size field of the OBU.
```
jwcullen, line 543: Minor typo "obu _size" has an extra space. "obu_size".

jwcullen (changes requested):
Approved after fixed typo. I think encoding the size this way is better.

jwcullen (approved):

index.bs
```diff
@@ -574,8 +574,6 @@ The reserved OBU allows the extension of this specification with additional OBU
 
 This section specifies obu payload of OBU_IA_Magic_Code.
 
-For this obu, the obu header (2 bytes) shall be set to 0xF807 (in case of [=obu_redundant_copy=] = 0) or FC07 (in case of [=obu_redundant_copy=] = 1) for this version of the specification.
```
yeroro, line 577: We can keep these sentences, just update to correct values:
```
0xF807 --> 0xF806
FC07   --> 0xFC06
```

yeroro (changes requested):

index.bs
sunghee-hwang: The values can be kept. But I am afraid that they may generate misunderstanding.

The values are correct for this version of the specification. But if we consider a future version with obu_extension_flag = 1, the values would be no longer correct.
That is the reason why I deleted the sentence.
Please consider it again.

yeroro: I see. Yeah I agree the Spec doesn't have to be unnecessarily restrictive.

yeroro (approved):

PR #329, opened 2023-03-23
Fix #274, updated ISOBMFF encapsulation (including trimming) and deca…

### LINKED ISSUE #274: Should we add a section describing how ISOBMFF fields need to be adjusted due to trimming?
If I understand, we have reached agreement that we would accept redundancy of trimming information in three places:  (1) within audio frame OBUs, (2) within the codec config describing Opus (when Opus is used), (3) using ELST and updating other timing information in ISOBMFF atoms.

I'm not aware of any spec that clearly documents all the moving parts that need to be updated in moov atoms, it's just implicitly described in ISOBMFF spec by the meaning of each field.

So, should we consider having an informative section that clarifies how the various ISOBMFF fields need to be adjusted to correctly represent trimming information?   The following fields may deserve being mentioned, though some of them the semantics would be that they are *unaffected* by the padding:

- ELST media time
- ELST segment duration
- mvhd duration
- mdhd duration
- sidx subsegment durations. (for fragmented mp4)
- tdft base media decode time (for fragmented mp4)
- trun last packet duration (and similar for non-fragmented)

Any other fields I may have missed?

sunghee-hwang:
…psulation

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/329.html" title="Last updated on Mar 28, 2023, 12:10 AM UTC (16d9d39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/329/ddaee5c...16d9d39.html" title="Last updated on Mar 28, 2023, 12:10 AM UTC (16d9d39)">Diff</a>


index.bs
```diff
@@ -2022,23 +2033,12 @@ This section provides a guideline for IAMF parser to reconstruct IA sequences fr
...
-	- Audio Element OBU(s): take the audio_element_obus in the ith AudioSampleEntry('iamf').
-	- Mix Presentation OBU(s): take the mix_presentation_obus in the ith AudioSampleEntry('iamf').
-- Step2: Reconstruction of Sync_OBU for the ith IA sequence.
-	- Take the sync_obu in the ith AudioSampleEntry('iamf').
-- Step3: Reconstructing of the jth Temporal Unit of the ith IA sequence (j = i1, i2, …, im)
-	- Take jth sample as it is.
-- Step4: Place descriptor OBUs, followed by Sync OBU, and followed by Temporal Units in order (j = i1, i2, …, im) without gap, to reconstruct the ith IA sequence.
+- The ith IA sequence is reconstructed as follows:
+	- Step1: Take the ith descriptor OBUs and Sync OBU from the ith IASampleEntry.
+	- Step2: Take jth sample as it is and add Temporal Delimiter OBU in front of the jth sample if necessary.
```
tdaede, line 2039: "if necessary" could be clarified to "if the temporal delimiter is not already present at the beginning of the sample". Though in theory it shouldn't be there anyway as the encapsulation process disallows it.

index.bs
```diff
@@ -1873,24 +1878,28 @@ IA sequence shall comply with the bitstream which is specified in [[#profiles-si
...
-
-- Each set of descriptor OBUs, specified in [[#standalone-descriptor-obus]], and the Sync OBU that immediately follows them are moved to one [=IASampleEntry=]. The descriptor OBUs are then discarded from the IA sequence. In addition,
-    - The [=num_samples_per_frame=] field in the Codec Config OBU is copied to 'stts'.
-    - The [=roll_distance=] field in the Codec Config OBU is copied to [=AudioRollRecoveryEntry=] with the [=grouping_type=] 'rol1'.
-- Each temporal unit:
+During encapsulation process, OBUs of IA sequences are encapsulated into [[!ISOBMFF]] as follows:
+
+- Movie Header and Edit
+	- Calculate the number of temporal units in the IA sequences.
+	- Calculate the number of audio samples to be trimmed at the start of the IA sequences and calculate the number of the audio samples to be trimmed at the end of the IA sequences.
```
tdaede, line 1885: I think we need to expand on this. What is the calculation method? Do we copy the values out as is of each substream? Use a more complex method?

index.bs
```diff
@@ -1873,24 +1878,28 @@ IA sequence shall comply with the bitstream which is specified in [[#profiles-si
...
+	- If there is audio samples to be trimmed at the start or at the end, then 'edts' and 'elst' boxes shall be present to reflect the the number of audio samples to be trimmed at the start of the IA sequence (i.e [=media_time=]) and [=edit_duration=].
+- Sample Entry
+	- The ith set of descriptor OBUs which is not a redundant copy of the previous descriptor OBUs (i.e. the first descriptor OBUs is regarded as non-redundant copy regardless of its [=obu_redundant_copy=]), specified in [[#standalone-descriptor-obus]], and the Sync OBU that immediately follows them are moved to the ith [=IASampleEntry=]. All descriptor OBUs in the IA sequences are then discarded from the IA sequence. 
+- Decoding Time to Sample
+	- The [=num_samples_per_frame=] value in the Codec Config OBU, which indicates the duration of each sample, is reflected to 'stts' box except the last sample. 
+	- The duration of the last sample may be the duration after trimming or before trimming. 
+	- The duration of teh last sample may not be same as [=num_samples_per_frame=] even it is the duration before trimming. In this case, it is calcuated from Audio Frame OBU of the last temporal unit.
```
tdaede, line 1893: the

index.bs
```diff
@@ -1873,24 +1878,28 @@ IA sequence shall comply with the bitstream which is specified in [[#profiles-si
...
+	- Reflect the calculated numbers to 'mvhd' to indicate the duration after trimming.
+	- If there is audio samples to be trimmed at the start or at the end, then 'edts' and 'elst' boxes shall be present to reflect the the number of audio samples to be trimmed at the start of the IA sequence (i.e [=media_time=]) and [=edit_duration=].
+- Sample Entry
+	- The ith set of descriptor OBUs which is not a redundant copy of the previous descriptor OBUs (i.e. the first descriptor OBUs is regarded as non-redundant copy regardless of its [=obu_redundant_copy=]), specified in [[#standalone-descriptor-obus]], and the Sync OBU that immediately follows them are moved to the ith [=IASampleEntry=]. All descriptor OBUs in the IA sequences are then discarded from the IA sequence. 
+- Decoding Time to Sample
+	- The [=num_samples_per_frame=] value in the Codec Config OBU, which indicates the duration of each sample, is reflected to 'stts' box except the last sample. 
+	- The duration of the last sample may be the duration after trimming or before trimming.
```
tdaede, line 1892: Not sure what this means. Is it allowing either option?

index.bs
```diff
@@ -1873,24 +1878,28 @@ IA sequence shall comply with the bitstream which is specified in [[#profiles-si
...
+- Sample Group
+	- The [=roll_distance=] value in the Codec Config OBU is copied to [=AudioRollRecoveryEntry=] with the [=grouping_type=] = 'roll'.
+- Sample
 	- Temporal Delimiter OBU: shall be discarded if present.	
-	- Remained OBUs of each temporal unit shall be stored as one sample data without gap among OBUs.
-- Audio Frame OBUs:
-	- Select one substream. 
-	- If [=obu_trimming_status_flag=] of the first Audio Frame OBU of the substream is set to 1, keep parsing following Audio Frame OBUs of the substream until meets the Audio Frame OBU having [=obu_trimming_status_flag=] =  0 and sum [=num_samples_to_trim_at_start=]. Then reflect the result of the summation to 'edts'.
-	- If [=obu_trimming_status_flag=] of the last Audio Frame OBU of the substream is set to 1, then reflect num_samples_to_trim_at_end to 'stts.
+	- Each temporal unit shall be stored as one sample data without gap among OBUs.
```
tdaede, line 1898: Specifically do you mean there shouldn't be any samples without any OBUs? Maybe to be clearer:

Each consecutive temporal unit should be placed in consecutive samples.

Unless you are intending to place a stronger constraint than that. (what if the input has a temporal unit that has no OBUs? I think that is effectively disallowed elsewhere but should the mapping process copy it as is, or specially reject it?)

tdaede:
I'm OK merging this once the comments are addressed. However, I'd really appreciate another review from someone very familiar with ISOBMFF.

index.bs
tdaede: "by taking the number of trimmed samples from the IA sequence and converting it to the sample rate used by the the 'elst' boxes."

index.bs
tdaede: "The duration of the last sample is duration including samples trimmed at the beginning but excluding samples trimmed at the end."

index.bs
tdaede: We can either remove "if necessary" to just always add a temporal delimiter OBU (because it should have been removed during encapsulation), or use the clarification above where we actually check if one is still there (which would have been a mistake during encapsulation).

tdaede (approved):

PR #331, opened 2023-03-24
Fix #330, to clarify definition of relative offset in sync OBU

### LINKED ISSUE #330: Need to clarify the meaning of relative_offset in Sync OBU
For example
- When relative offset = 0 for audio_substream_id = 0 and relative offset = 0 for parameter_id = 0
It is unclear whether the parameter data of the parameter block starts to be applied to the first audio sample before trimming or after trimming.

I think that it should start to be applied to the first audio sample after trimming of the audio frame.
In above example, if the audio_substream_Id has num_samples_to_trim_at_start, then relative offset for the parameter_id should be able to let for decoders know the location of the audio sample where the first parameter data is applied to, at any random access point.

So, the difference (i.e. relative_offset (for audio_substream_id) - relative_offset (for parameter_id)) = num_samples_to_trim_at_start.

Also, in this point, 
[As Is]
```
The offset shall be indicated in the number of ticks at the `parameter_rate` specified in the corresponding substream or parameter definition.
```
[To Be]
```
The offset shall be indicated in the number of ticks at the 'sample_rate' specified in the corresponding substream or Codec Config OBU. For IAMF-OPUS, the `sample_rate` is 48000Hz.
```

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/331.html" title="Last updated on Mar 27, 2023, 11:10 PM UTC (c7fd867)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/331/ddaee5c...c7fd867.html" title="Last updated on Mar 27, 2023, 11:10 PM UTC (c7fd867)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1100,7 +1100,10 @@ For this version of the specification, the value of reinitialize_decoder shall b
 
 <dfn noexport>relative_offset</dfn> is the offset to position the first audio frame (before trimming) or parameter block with the referenced obu_id that comes after this Sync OBU with respect to the timeline generated before this Sync OBU. If this Sync OBU is the first one, it is the offset from 0. Otherwise, it is the offset from the end of the timeline of Substreams generated from the previous Sync OBU.
 
-The offset shall be indicated in the number of ticks at the [=parameter_rate=] specified in the corresponding substream or parameter definition.
+The difference, [=relative_offset=](P) - [=relative_offset=](A), shall be the offset to position the first audio sample where the first parameter value of the first parameter block that comes after this Sync OBU is applied to.
+- Where, [=relative_offset=](P) is [=relative_offset=] of [=obu_id=] for the parameter and [=relative_offset=](A) is [=relative_offset=] of [=obu_id=] for the substream which the parameter is applied to. 
+
+The offset shall be indicated in the number of ticks at the sample rate specified in the corresponding substream or Codec Config OBU. For IAMF-OPUS, the sample rate is 48000Hz.
 
 IA encoder and decoder operations related to this field are specified in [[#standalone-synchronizing-data-obus]].
 


yeroro (approved):

yeroro:
I just have one question: what if the `parameter_rate` is different from the audio codec's `sample_rate`? Does the `relative_ offset` use different tick sizes for audio vs parameters? Or should we resample the parameters to 48 kHz like when audio samples are in different sample rates (as in 9.3.3. Mixing Audio Elements)?

sunghee-hwang:
> I just have one question: what if the `parameter_rate` is different from the audio codec's `sample_rate`? Does the `relative_ offset` use different tick sizes for audio vs parameters? Or should we resample the parameters to 48 kHz like when audio samples are in different sample rates (as in 9.3.3. Mixing Audio Elements)?

In old vesion, the units of relative offset for parameter_id were based on its parameter_rate. But, this PR intends to change it to the sample_rate of the substream which the parameter id is applied to.

FYI, parameter_rate is the sampling rate of parameter values. If parameter_rate differs from audio sample rate of the substream which the parameter is applied to, then I think that the parameter values need to be resampled to the audio sample rate.  
A good example is the text in Section 9.4 Animated Parameters
```
To apply the parameter values to the audio samples in the segment without interpolation, the [parameter_rate] is first resampled to the audio sample rate to give:

n0 = t0 * audio_sample_rate / parameter_rate,
n1 = t1 * audio_sample_rate / parameter_rate,
n2 = t2 * audio_sample_rate / parameter_rate.
```

tdaede:
Whoops, @yeroro did you mean to close this?

PR #351, opened 2023-04-05
Fix #95, updated introduction

### LINKED ISSUE #95: Improve the introduction
The introduction could be improved:
- Some context is missing before defining what is "IA bitstream".
- It overlaps with section 3 on the Overview

I suggest restructuring as follows:
* start by defining what "immersive audio" means, for example, "the combination of 3D audio signals recreating a sound experience close to that of a natural environment"
* then indicates that this specification defines a model for representing immersive audio content based on coded audio sub-streams contributing to audio elements meant to be rendered and mixed to form one or more immersive presentations (Figure 2)
*  then indicates that this specification defines a hypothetical architecture based on pre-processing of the source signal to split the signal into substreams and determine metadata to drive the rendering and mixing, then based on separate encoding of the substreams and the metadata, and then on the combination of these to form a bitstream. You can reuse Figure 1 and the bullet list below, but with some updates (should not mention IAC file)
* then indicates that this specification defines a way to store the bitstream into container formats for applications that need it

I don't think we need to list the rest of the specification. This is error prone and the specification already contains a table of content on the side.

As a result, section 1 and 2 should be merged and only 1 section "Introduction" should be used.

sunghee-hwang:
Please refer to that ID linking scheme is moved to Annex.

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/351.html" title="Last updated on Apr 5, 2023, 1:57 AM UTC (45cbb3f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/351/fb04c27...45cbb3f.html" title="Last updated on Apr 5, 2023, 1:57 AM UTC (45cbb3f)">Diff</a>


tdaede:
I'll review this today editorially and then will merge when r+'d.

tdaede (approved):

PR #361, opened 2023-04-12
Fix #353, segment replaced by subblock

### LINKED ISSUE #353: Suggest renaming parameter "segment" and "segment_interval"
The word segment may become overloaded with the idea of DASH segments.

Would it make sense to call it a subblock instead?   i.e. a parameter block may have 1 or more subblocks that define actual values of the parameter over time.

And further, instead of "segment_interval" -> this really is "duration", I feel that interval would be two values describing a beginning point and end point, with questions about semantics of open/closed boundaries.   So would it make sense to rename this to "subblock_duration"?

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/361.html" title="Last updated on Apr 13, 2023, 10:46 PM UTC (440e49f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/361/537a92e...440e49f.html" title="Last updated on Apr 13, 2023, 10:46 PM UTC (440e49f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -720,11 +720,11 @@ abstract class ParamDefinition() {
   unsigned int (7) reserved;
   if (param_definition_mode == 0) {
     leb128() duration;
-    leb128() num_segments;
-    leb128() constant_segment_interval;
-    if (constant_segment_interval == 0) {
-      for (i=0; i< num_segments; i++) {
-        leb128() segment_interval;
+    leb128() num_subblocks;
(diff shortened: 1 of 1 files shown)


yeroro (commented):
Shawn also suggested using "_duration" instead of "_interval". Do you also agree?

yeroro (approved):

PR #366, opened 2023-04-18
Fix #365, clarified sample format for every sample to be sync sample

### LINKED ISSUE #365: Need to clarify the definition of IA sample format for every sample to be sync sample
The current spec says followings but unclear if every sample is sync sample.

6.2.5. IA Sample Format
For tracks using the [IASampleEntry](https://aomediacodec.github.io/iamf/#iasampleentry)(s), an IA Sample has the following constraints:

The one sample data shall be one temporal unit after removing Temporal Delimiter OBU if present.

A temporal unit consists of all the data OBUs that are associated with a specific, distinct time instant. It consists of zero or one Temporal Delimiter OBU, and followed by one or more Audio Frame OBUs and zero or more Parameter Block OBUs.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/366.html" title="Last updated on Apr 25, 2023, 1:20 AM UTC (5169b20)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/366/c9595f3...5169b20.html" title="Last updated on Apr 25, 2023, 1:20 AM UTC (5169b20)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1942,10 +1942,10 @@ class IA_Descriptor extends Box('iamd') {
 For tracks using the [=IASampleEntry=](s), an <dfn noexport>IA Sample</dfn> has the following constraints:
 - The one IA Sample data shall be N (N > 0) consecutive temporal units (after removing Temporal Delimiter OBU if present) for every IA Sample to be sync sample. And, every IA Sample except the last IA Sample shall have the same number of temporal units.
 	- In other words, IA Sample consists of all the data OBUs that are associated with a specific, distinct time instant so that it consists of N consecutive Audio Frame OBUs per substream and one or more Parameter Block OBUs per parameter (if present), whose total duration is same as that of the N consecutive Audio Frame OBUs. 
-- In every IA Sample, the relative offset of each substream or each parameter (if present) is idenfied by its associated [=relative_offset=] field of the sync_obu() in [=IA_Descriptor=].
+- In every IA Sample, the relative offset of each substream or each parameter (if present) is identified by its associated [=relative_offset=] field of the sync_obu() in [=IA_Descriptor=].
 	- In other words, [=relative_offset=](s) in sync_obu() is (are) applicable to every IA Sample in its associated [=IASampleEntry=].
 
-A temporal unit consists of zero or one Temporal Delimiter OBU, and followed by one Audio Frame OBU per substream and zero or one one Parameter Block OBU per parameter.
+A temporal unit consists of zero or one Temporal Delimiter OBU, and followed by one Audio Frame OBU per substream and zero or one Parameter Block OBU per parameter.
 
 ## Codecs Parameter String ## {#codecsparameter}
 DASH and other applications require defined values for the 'Codecs' parameter specified in [[!RFC6381]] for ISO Media tracks. The codecs parameter string for the AOM IA codec shall be:
--- a/index.bs
+++ b/index.bs
@@ -1941,11 +1941,9 @@ class IA_Descriptor extends Box('iamd') {
 
 For tracks using the [=IASampleEntry=](s), an <dfn noexport>IA Sample</dfn> has the following constraints:
 - The one IA Sample data shall be N (N > 0) consecutive temporal units (after removing Temporal Delimiter OBU if present) for every IA Sample to be sync sample. And, every IA Sample except the last IA Sample shall have the same number of temporal units.
-	- In other words, IA Sample consists of all the data OBUs that are associated with a specific, distinct time instant so that it consists of N consecutive Audio Frame OBUs per substream and one or more Parameter Block OBUs per parameter (if present), whose total duration is same as that of the N consecutive Audio Frame OBUs. 
+	- In other words, every IA Sample except the last IA Sample consists of all the data OBUs that are associated with a specific, distinct time instant so that it consists of N consecutive Audio Frame OBUs per substream and one or more Parameter Block OBUs per parameter (if present), the total duration of the one or more Parameter Block OBUs is equal to the total duration of the remaining Audio Frame OBU(s) of the N consecutive Audio Frame OBUs, excluding the Audio Frame OBU(s) for which the entire audio frame is trimmed. 
 - In every IA Sample, the relative offset of each substream or each parameter (if present) is identified by its associated [=relative_offset=] field of the sync_obu() in [=IA_Descriptor=].
 	- In other words, [=relative_offset=](s) in sync_obu() is (are) applicable to every IA Sample in its associated [=IASampleEntry=].
-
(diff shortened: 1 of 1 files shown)


yeroro:
Can someone with more knowledge of MP4 review this? I am afraid I cannot.

index.bs
```diff
@@ -1940,9 +1944,17 @@ class IA_Descriptor extends Box('iamd') {
 ### IA Sample Format ### {#iasampleformat}
 
 For tracks using the [=IASampleEntry=](s), an <dfn noexport>IA Sample</dfn> has the following constraints:
-- The one sample data shall be one temporal unit after removing Temporal Delimiter OBU if present.
-
-A temporal unit consists of all the data OBUs that are associated with a specific, distinct time instant. It consists of zero or one Temporal Delimiter OBU, and followed by one or more Audio Frame OBUs and zero or more Parameter Block OBUs.
+- One IA Sample data shall be one or more consecutive temporal units (after removing Temporal Delimiter OBUs if present) for every IA Sample to be sync sample.
```
tdaede, line 1947: I think this should still be as before, one temporal unit == one IA sample.

index.bs
```diff
@@ -1940,9 +1944,17 @@ class IA_Descriptor extends Box('iamd') {
...
+	- In other words, every IA Sample consists of all the data OBUs that are associated with a specific, distinct time instant.
+- If [=num_samples_to_trim_at_start=] < [=num_samples_per_frame=], every IA Sample except the last IA Sample shall have the same number (N > 0) of temporal units.
```
tdaede, line 1949: Basically even if we trim more than entire audio frame (or even an entire temporal unit) we should still have it be a distinct IA sample. It will still be trimmed by the edit list information as normal.

tdaede (changes requested):

index.bs
tdaede: So this allows multiple temporal units per IA sample in all cases.

index.bs
```diff
@@ -1855,16 +1859,16 @@ During encapsulation process, OBUs of IA sequences are encapsulated into [[!ISOB
...
 - Sample Entry
 	- The ith set of descriptor OBUs which is not a redundant copy of the previous descriptor OBUs (i.e. the first descriptor OBUs is regarded as non-redundant copy regardless of its [=obu_redundant_copy=]), specified in [[#standalone-descriptor-obus]], and the Sync OBU that immediately follows them are moved to the ith [=IASampleEntry=]. All descriptor OBUs in the IA sequences are then discarded from the IA sequence. 
-- Decoding Time to Sample
-	- The [=num_samples_per_frame=] value in the Codec Config OBU, which indicates the duration of each sample, is reflected in the 'stts' or 'trun' box except the last sample. 
-	- The duration of the last sample is duration including samples trimmed at the beginning but excluding samples trimmed at the end. 
-	- The duration of the last sample may not be same as [=num_samples_per_frame=]. In this case, it is calcuated from Audio Frame OBU of the last temporal unit.
+- Decoding Time to IA Sample
+	- The [=num_samples_per_frame=] value in the Codec Config OBU, which indicates the duration of one audio frame, is reflected in the 'stts' or 'trun' box based on [=IA Sample=] except the last [=IA Sample=].
```
tdaede, line 1863: If we allow more than one temporal unit / audio frame per IA sample, then num_samples_per_frame should match the duration of the IA sample.

tdaede:
This design means that temporal blocks no map 1:1 to IA Samples. This means that the encoder has more freedom to map, but also means that when temporal units are stripped during packaging, they will not be re-created when demuxing. I need to review this deeper to see if it will work correctly.

Alternately, either we could leave it as is with a 1:1 mapping, and do a simpler PR that simply limits parameter blocks such that a 1:1 mapping works with every sample being sync samples, or allow multiple audio frames inside a single temporal unit. @shawnisingh will take a look at this in more detail as well.

tdaede:
I'll also review this for any technical blockers, assuming that we go with this method, now that I better understand what is being achieved.

index.bs
sunghee-hwang: > This design means that temporal blocks no map 1:1 to IA Samples. This means that the encoder has more freedom to map, but also means that when temporal units are stripped during packaging, they will not be re-created when demuxing. I need to review this deeper to see if it will work correctly.
> 
> Alternately, either we could leave it as is with a 1:1 mapping, and do a simpler PR that simply limits parameter blocks such that a 1:1 mapping works with every sample being sync samples, or allow multiple audio frames inside a single temporal unit. @shawnisingh will take a look at this in more detail as well.

If no one has concern, then I would prefer one temporal unit = one audio frame = one IA sample with every sample being sync samples for v1. (For this, we need to limit parameter blocks such that the duration of parameter block = the duration of audio frame).
Instead of that, one IA sample = multiple temporal unit (one temporal unit = one audio frame) or IA sample = one temporal unit with multiple audio frames is moved to v2.

index.bs
```diff
@@ -1860,9 +1860,8 @@ During encapsulation process, OBUs of IA sequences are encapsulated into [[!ISOB
 - Sample Entry
 	- The ith set of descriptor OBUs which is not a redundant copy of the previous descriptor OBUs (i.e. the first descriptor OBUs is regarded as non-redundant copy regardless of its [=obu_redundant_copy=]), specified in [[#standalone-descriptor-obus]], and the Sync OBU that immediately follows them are moved to the ith [=IASampleEntry=]. All descriptor OBUs in the IA sequences are then discarded from the IA sequence. 
 - Decoding Time to IA Sample
-	- The [=num_samples_per_frame=] value in the Codec Config OBU, which indicates the duration of one audio frame, is reflected in the 'stts' or 'trun' box based on [=IA Sample=] except the last [=IA Sample=]. 
+	- The duratin of [=IA Sample=], which is reflected in 'stts' or 'trun' box, is calculated based on the number of audio frames which [=IA Sample=] includes and the [=num_samples_per_frame=] value in the Codec Config OBU, which indicates the duration of one audio frame.
```
tdaede, line 1863: small typo: duration

tdaede:
I reviewed how the spec would operate with this current PR with several temporal units per sample and my main concern is that when we unpack this back to a bitstream, any temporal OBUs added are now only added per sample, e.g. it does not round trip. I tried to find places where making the temporal units less frequent would break things and found the following text:

> Every Temporal Unit shall have Temporal Delimiter OBU or no Temporal Unit shall have Temporal Delimiter OBU.

I suppose the demuxer could parse the bitstream and add back multiple temporal delimiter OBUs based on reading audio frame OBUs (or just not output them... it does make temporal delimiter OBUs seem a bit vestigal if it is easy to determine where they should go just by parsing the other OBUs)

tdaede (approved):
This gets an approval from me because I believe it technically doesn't break anything because you are always allowed to demux to a stream without temporal delimiters, but I think that issue should be addressed in more detail (or we go with 1:1 mapping)

PR #374, opened 2023-04-24
359: Add anchored loudness and associated fields to bitstream spec

phastphill:


index.bs
```diff
@@ -1231,6 +1237,14 @@ class loudness_info() {
   if (info_type & 1) {
     signed int (16) true_peak;
   }
+
+  if (info_type & 2) {
+    leb128() num_anchored_loudness;
```
phastphill, line 1242: Since `anchor_element` is 8 bits, `num_anchored_loudness` could also be an 8-bit parameter

index.bs
sunghee-hwang, line 1: Needs to be 127 to 255

index.bs
```diff
@@ -1241,15 +1255,27 @@ class loudness_info() {
...
 
 <dfn noexport>true_peak</dfn> specifies the true peak of the audio signal, specified in dBFS and measured according to [[!ITU1770-4]].
 
+<dfn noexport>anchor_element</dfn> specifies the anchor element used in computation of the anchored_loudness which follows, as defined in [[!ISOIEC-23091-3-2018]], as follows:
+
+<pre class = "def">
+   0   : Unknown
+   1   : Dialogue
+   2   : Album
+ 3~127 : Reserved
```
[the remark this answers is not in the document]
phastphill: Thanks

tdaede (changes requested):
It looks like I can repeat arbitrarily many loudnesses. What happens if I specify 30 anchored loudnesses (or true peaks)?

phastphill:
@tdaede do you mean that there could be multiple loudness_info blocks in a stream?  If that's the issue, could we add some text around how often a loudness OBU can appear?  i.e. there must be at least one frame of audio data in between loudness_info blocks.

tdaede:
No, I mean because this is currently a list of {type, loudness} pairs, you can place duplicate types in the stream. That should be disallowed and/or the decoder should have a rule of which value it picks (e.g. the last)

phastphill:
Okay, I understand.  Yes, it is possible to put duplicates into the field.  We should rule it out if at all possible.  Is there a mechanism like this elsewhere in the spec?

index.bs
```diff
@@ -1241,15 +1255,31 @@ class loudness_info() {
...
+<dfn noexport>anchor_element</dfn> specifies the anchor element used in computation of the anchored_loudness which follows, as defined in [[!ISOIEC-23091-3-2018]], as follows:
+
+<pre class = "def">
+   0   : Unknown
+   1   : Dialogue
+   2   : Album
+ 3~255 : Reserved
+</pre>
+
+There shall be no duplicate values of [=anchor_element=] within one loudness_info() OBU.
```
jwcullen, line 1277: Drop OBU here.

jwcullen (changes requested):
Sorry one last minor change because the terminology is slightly wrong and confusing here. Just drop OBU in the recently added sentence.

_There shall be no duplicate values of [=anchor_element=] within one loudness_info()._

There is no loudness_info() OBU. The OBU here is the Mix Presentation OBU. Which is allowed to have multiple submixes. Each submix has a single loudness_info(). A use case could be a single Mix Presentation describes an English-audio submix and and an Italian-audio submix. Each would have its own loudness_info() struct which may truly have unique values, so I think dropping OBU here is correct and captures the uniqueness limitation we were aiming for.

Once that is fixed it looks like all issues brought up are resolved. I think it is good to go if there are no further objects

PR #379, opened 2023-04-25
Fix #377, substream coding and order are moved to OBU syntax

### LINKED ISSUE #377: Exact channel couplings are not normatively required
The generation process currently suggests the following channel couplings:

        Coupled channels
        : L/R, Ls/Rs, Lss/Rss, Lrs/Rrs, Ltf/Rtf, Ltb/Rtb
        Non-coupled channels: C, LFE, L 

However it does not mandate them elsewhere in the spec. If we want this to be disallowed then we should be more explicit about it in a normative section.

Test vector 3 in https://github.com/AOMediaCodec/libiamf/pull/3 currently uses non coupled channels.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/379.html" title="Last updated on Apr 26, 2023, 1:46 AM UTC (3fa656a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/379/3036657...3fa656a.html" title="Last updated on Apr 26, 2023, 1:46 AM UTC (3fa656a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2436,7 +2436,7 @@ Where, p1 = 0.707. Implementations may use limiter defined in [[#processing-post
 
 # IAMF Generation Process # {#iamfgeneration}
 
-This section provides a process for IA encoding for a given input audio format.
+This section provides a guideline for IA encoding for a given input audio format.
 
 Recommended input audio format for IA encoding is as follows:
 - Ambisonics format: It shall conform to [=ChannelMappingFamily=] = 2 or 3 of [[RFC8486]].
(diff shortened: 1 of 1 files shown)


sunghee-hwang (commented):
the word "guideline" will be changed to "process" in another PR. So, it backs to the original,

tdaede:
I'd like @yeroro to take a look at this as it's relevant to [libiamf.](https://github.com/AOMediaCodec/libiamf/issues/5)

PR #395, opened 2023-05-11
Fix #389, only the last frame is allowed for the trimming at end

### LINKED ISSUE #389: More than one frame of audio samples can be trimmed at the end
For obu_status_trimming flag, the spec says:

> Thus, [obu_trimming_status_flag](https://aomediacodec.github.io/iamf/#obu_trimming_status_flag) shall be set to 1 only for one or more consecutive Audio Frame OBUs at the start of the substream and/or only for one or more consecutive Audio Frame OBUs at the end of the substream.

We could restrict this to only one Audio Frame for the end of the spec.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/395.html" title="Last updated on May 22, 2023, 12:21 AM UTC (8893820)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/395/f19474f...8893820.html" title="Last updated on May 22, 2023, 12:21 AM UTC (8893820)">Diff</a>


index.bs
```diff
@@ -510,7 +510,7 @@ It shall always be set to 0 for the following obu_type values:
 <dfn noexport>obu_trimming_status_flag</dfn> indicates whether this OBU has audio samples to be trimmed or not. If it is set to 0, the [=num_samples_to_trim_at_start=] and [=num_samples_to_trim_at_end=] fields shall not be present. Otherwise, the [=num_samples_to_trim_at_start=] and [=num_samples_to_trim_at_end=] fields shall be present.
```
cconcolato, line 510: The use of 'shall' in this sentence is not needed. It is a statement of fact that `num_samples_to_trim_at_start` and `num_samples_to_trim_at_end` can only be coded when the `obu_trimming_status_flag` is set. I suggest removing the sentences as it is covered in the syntax. On the other hand, it should be clear that it can only be used when the corresponding OBU is an Audio OBU.
``` suggestion
<dfn noexport>obu_trimming_status_flag</dfn> indicates whether this OBU has audio samples to be trimmed or not. It SHALL be set only when [=obu_type=] is set to `OBU_IA_Audio_Frame` or `OBU_IA_Audio_Frame_ID0` to `OBU_IA_Audio_Frame_ID21`.
```

index.bs
```diff
@@ -510,7 +510,7 @@ It shall always be set to 0 for the following obu_type values:
...
 
 For a given substream, 
-- If the substream contains audio samples to be trimmed, then they shall only be at the start and/or end of the substream. Thus, [=obu_trimming_status_flag=] shall be set to 1 only for one or more consecutive Audio Frame OBUs at the start of the substream and/or only for one or more consecutive Audio Frame OBUs at the end of the substream.
+- If the substream contains audio samples to be trimmed, then they shall only be at the start and/or end of the substream. Thus, [=obu_trimming_status_flag=] shall be set to 1 only for one or more consecutive Audio Frame OBUs at the start of the substream and/or only for one Audio Frame OBU at the end of the substream.
```
cconcolato, line 513: This raises several questions:
- What is the start of a substream? Is it the Audio OBUs with a give substream id immediately following the Codec OBU for the same substream id?
- What is the end of a substream? Is that the last Audio Frame OBUs for a given substream id before any non-redundant Codec OBU reusing the same substream id?
- What should be the value of `num_samples_to_trim_at_end` for the Audio Frame OBUs at the start of the substream?
- What should be the value of `num_samples_to_trim_at_start` for the Audio Frame OBUs at the end of the substream?

The sentence:
> If the substream contains audio samples to be trimmed, then they shall only be at the start and/or end of the substream.

uses a 'shall' that would be hard to verify by a validator. How would a validator know that the 'If' part is applicable? 
```suggestion
- If an Audio Frame OBU has its [=num_samples_to_trim_at_start=] field set to a non-zero value N, the decoder SHALL discard the first N audio samples.
- If an Audio Frame OBU has its [=num_samples_to_trim_at_end=] field set to a non-zero value N, the decoder SHALL discard the last N audio samples.
NOTE: Because of coding dependency, discarding a sample can sometimes mean decoding the entire audio frame.
- For a given Audio Frame OBU, the sum of [= num_samples_to_trim_at_start=] and [= num_samples_to_trim_at_end=] shall be less or equal to the number of samples in the Audio Frame OBU. 
NOTE: This means that if one of the value is set to the the number of samples in the Audio Frame OBU, the other value is set to 0.
- When [=num_samples_to_trim_at_start=] is non-zero, all Audio Frame OBUs with the same substream id, and preceding this OBU back until the Codec OBU defining this substream, SHALL have their [=num_samples_to_trim_at_start=] field equal to the number of samples in the corresponding Audio Frame OBU.
- When [=num_samples_to_trim_at_end=] is non-zero in an Audio Frame OBU, there shall be no subsequent Audio Frame OBU with the same substream id until a non-redundant Codec OBU defining a substream with the same substream id.
```

index.bs
```diff
@@ -510,7 +510,7 @@ It shall always be set to 0 for the following obu_type values:
...
 - Otherwise, then [=obu_trimming_status_flag=] shall be set to 0 for every Audio Frame OBU of the substream.
```
cconcolato, line 514: Given the above, I would suggest deleting this sentence
```suggestion
```

index.bs
sunghee-hwang: @cconcolato,

I have a difficult to get the meaning of the sentence in below.

- When [=num_samples_to_trim_at_start=] is non-zero, all Audio Frame OBUs with the same substream id, and preceding this OBU back until the Codec OBU defining this substream, SHALL have their [=num_samples_to_trim_at_start=] field equal to the number of samples in the corresponding Audio Frame OBU.

Does it mean the Audio Frame OBUs which are trimmed entirely?

sunghee-hwang: @cconcolato, thank you making me remind it.

Here are my opinions.
- a Substream consists of a series of audio packets ordered in time which are generated by a given encoder without re-initialization.
- a Substream has an unique `audio_stream_id`. So, each audio frame obu storing each audio packet has the `audio_stream_id`.
- an Audio Element OBU indicates the `audio_stream_id` which the Audio Element refers to.
- When the lifetime of the Audio Element OBU is ended, the lifetime of the Substream is ended.
- We need to discuss when the lifetime of an Audio Element OBU is ended. (Not specified in our specification) 
   .IMHO, the lifetime of an Audio Element OBU starts from non-redundant Audio Element OBU with an `audio_element_id` and ends when the following Descriptor OBUs do not have Audio Element OBU with the `audio_element_id` or when the following Descriptor OBUs do have non-redundant Audio Element OBU with the `audio_element_id` or when there are no following Descriptor OBUs. 

Based on this, I would define the start and the end of a substream like this.
- the start of a Substream is the first Audio Frame OBUs among all Audio Frame OBUs with the same `audio_substream_id` immediately following a non-redundant Audio Element OBU using the `audio_substream_id` and (if present) before the next non-redundant Audio Element OBU reusing the same `audio_substream_id`.
- the end of a Substream is the last Audio Frame OBU for the `audio_substream_id` before any non-redundant Audio Element OBU reusing the same `audio_substream_id`.

cconcolato: > Does it mean the Audio Frame OBUs which are trimmed entirely?

No. The suggested sentence means there is no Audio Frame OBU at all for this substream after a end-trimmed Audio Frame OBU.

cconcolato: > the start of a Substream is the first Audio Frame OBUs among all Audio Frame OBUs with the same audio_substream_id immediately following a non-redundant Audio Element OBU using the audio_substream_id and (if present) before the next non-redundant Audio Element OBU reusing the same audio_substream_id.

I agree but I think it can be simplified. I think the sentence is saying twice the same thing: the first (in time order) Audio Frame OBU with a given `audio_substream_id` is necessarily the first (in byte order), i.e. the one following the Audio Element OBU. I would therefore say:
> the start of a Substream is the first Audio Frame OBUs in time order among all Audio Frame OBUs with the same audio_substream_id. It follows immediately the non-redundant Audio Element OBU using the audio_substream_id. If a non-redundant Audio Element OBU reusing the same audio_substream_id is used, the start of the substream is before that OBU.

> the end of a Substream is the last Audio Frame OBU for the audio_substream_id before any non-redundant Audio Element OBU reusing the same audio_substream_id.

I'm not sure. This means that a substream can span across non-redundant Audio Element OBU (not using the audio stream id), which means that all substreams don't have the same start and end. This makes things more complex, but maybe there are use cases. If you choose to keep it, I would suggest adding an explicit note saying that not all substreams start and end at the same time and that profiles may restrict that. I would suggest considering adding this restriction in the simple profile.

sunghee-hwang: > > Does it mean the Audio Frame OBUs which are trimmed entirely?
> 
> No. The suggested sentence means there is no Audio Frame OBU at all for this substream after a end-trimmed Audio Frame OBU.

Thanks for the clarification. Now, I get your intention.

cconcolato: Sorry. You were right from the beginning. The sentence was meant to say that if any Audio Frame OBU uses `num_samples_to_trim_at_start`, it means all Audio Frame OBUs before (if any) have to be entirely trimmed.

tdaede:
@cconcolato can you approve if all issues are addressed?

cconcolato (approved):

tdaede:
Needs HTML removed before merge.

sunghee-hwang:
It seems to me that the sub-branch issue_389 has not haven HTML file. I have no idea why it is there. It would be better to delete HTML if present after merging it.

cconcolato:
I think we're good to merge. The PR actually deletes the HTML (see the `-` sign next to the file name in the "Files Changed" tab).

PR #424, opened 2023-05-22
Fix #398, Magic Code OBU Improvement

### LINKED ISSUE #398: Magic Code OBU improvements
While reviewing the section about Magic Code OBU, I think a few improvements could be done:
- It is not clear what the purpose of the OBU is used for, why it is necessary. I suggesting adding a note indicating that this magic code OBU is added to permit file type sniffing (https://mimesniff.spec.whatwg.org/)
- It is not clear how Magic Code OBUs shall be treated when packaged in container formats. I suggest adding a note that Magic Code OBU MAY be removed when stored in formats that already provide type identification.
- The definition of the 4CC says "ia_code indicates a ‘four-character code’ (4CC) to identify the start of the IA sequence. It shall be ‘iamf’." The part about starting the sequence is already covered in the Standalone Format definition.
- Add a note that Magic Code OBU shall not be set as redundant. In other words, if a new Magic Code OBU appears in a sequence of OBUs, it starts a new IA Sequence. This could happen when concatenating 2 IA Sequences together.
- The `version` and `profile_version` fields are on 8 bits, while they really on 2 4-bit fields. Consider splitting them. Also, generally speaking it is not clear how to use them, what the consequences on readers. Rephrase semantics with clear shall/should for sequence and for implementations.

sunghee-hwang:
Comments for version and profile_version need to be addressed in another issue.

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/424.html" title="Last updated on May 25, 2023, 10:21 PM UTC (efe222c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/424/d1b7b59...efe222c.html" title="Last updated on May 25, 2023, 10:21 PM UTC (efe222c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -555,6 +555,8 @@ This section specifies obu payload of OBU_IA_Magic_Code.
 
 This OBU is used to indicate the start of IA Sequence. So, the first OBU of IA Sequence shall be OBU_IA_Magic_Code.
 
+NOTE: When an IA Sequence is stored in a file, the Magic Code OBU can be used to detect that the file contains an IA Sequence.
+
 This OBU may be placed frequently within one single IA Sequence for an application such as broadcasting or multicasting scenario. In that case, the other Magic Code OBUs except the first one shall be marked as redundant (i.e. [=obu_redundant_copy=] = 1).
 
 <b>Syntax</b>
--- a/index.bs
+++ b/index.bs
@@ -551,7 +551,13 @@ The reserved OBU allows the extension of this specification with additional OBU 
 
 ## Magic Code OBU Syntax and Semantics ## {#obu-magiccode}
 
-This section specifies obu payload of OBU_IA_Magic_Code.
+This section specifies obu payload of OBU_IA_Magic_Code. 
+
+This OBU is used to indicate the start of IA Sequence. So, the first OBU of IA Sequence shall be OBU_IA_Magic_Code.
+
+This OBU may be placed frequently within one single IA Sequence for an application such as broadcasting or multicasting scenario. In that case, the other Magic Code OBUs except the first one shall be marked as redundant (i.e. [=obu_redundant_copy=] = 1).
+
+NOTE: Non-redundant Magic Code OBU (i.e. [=obu_redundant_copy=] = 0) coming after the first Magic Code OBU is regarded as the start of a new IA sequence. This could happen when concatenating 2 IA Sequences together.
 
 <b>Syntax</b>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -551,7 +551,13 @@ The reserved OBU allows the extension of this specification with additional OBU
...
 ## Magic Code OBU Syntax and Semantics ## {#obu-magiccode}
 
-This section specifies obu payload of OBU_IA_Magic_Code.
+This section specifies obu payload of OBU_IA_Magic_Code. 
+
+This OBU is used to indicate the start of IA Sequence. So, the first OBU of IA Sequence shall be OBU_IA_Magic_Code.
+
+This OBU may be placed frequently within one single IA Sequence for an application such as broadcasting or multicasting scenario. In that case, the other Magic Code OBUs except the first one shall be marked as redundant (i.e. [=obu_redundant_copy=] = 1).
+
+NOTE: Non-redundant Magic Code OBU (i.e. [=obu_redundant_copy=] = 0) coming after the first Magic Code OBU is regarded as the start of a new IA sequence. This could happen when concatenating 2 IA Sequences together.
```
tdaede, line 560: If we define streams that have multiple IA sequences in series, this implies the decoder has to support it (and we would need reference streams in libiamf).

I would suggest removing this NOTE: section for now and only define what happens for a single IA sequence. @cconcolato does that sound reasonable?

index.bs
```diff
@@ -551,7 +551,11 @@ The reserved OBU allows the extension of this specification with additional OBU
 
 ## Magic Code OBU Syntax and Semantics ## {#obu-magiccode}
 
-This section specifies obu payload of OBU_IA_Magic_Code.
+This section specifies obu payload of OBU_IA_Magic_Code. 
+
+This OBU is used to indicate the start of IA Sequence. So, the first OBU of IA Sequence shall be OBU_IA_Magic_Code.
```
cconcolato, line 556: We could add a NOTE as follows:
```
NOTE: When an IA Sequence is stored in a file, the Magic Code OBU can be used to detect that the file contains an IA Sequence.
```

index.bs
```diff
@@ -565,7 +569,9 @@ class magic_code_obu() {
 
 <b>Semantics</b>
 
-<dfn noexport>ia_code</dfn> indicates a ‘four-character code’ (4CC) to identify the start of the IA sequence. It shall be ‘iamf’.
+<dfn noexport>ia_code</dfn> indicates a ‘four-character code’ (4CC), ‘iamf’. The OBU header for this OBU and this field are used as in-band signal (i.e. sync pattern) to indicate the start of IA Sequence.
+
+NOTE: For example, when [=obu_extention_flag=] = 0, the value of this OBU header is 0xF806 or 0xFC06. So, the sync pattern is 0xF806 + ‘iamf’ or 0xFC06 + ‘iamf’.
```
cconcolato, line 574:
```suggestion
<dfn noexport>ia_code</dfn> indicates a ‘four-character code’ (4CC), ‘iamf’. 
NOTE: When IA OBUs are delivered over a protocol that does not provide explicit IA Sequence boundaries, a parser may locate the IA Sequence start by searching for the code 'iamf' preceded by specific OBU header values, e.g. assuming [=obu_extension_flag=] is set to 0 and because [=obu_trimming_status_flag=] is set to 0 for a Magic Code OBU, the OBU header can be 0xF806 or 0xFC06
```

cconcolato (changes requested):

index.bs
cconcolato: I think we should be explicit in the specification:
- The model part of the only defines a single IA Sequence. The IA Sequence starts with the first Magic Code OBU (typically non-redundant) and ends with the next non-redundant Magic Code OBU. There are consequences, e.g. Descriptor OBUs shall follow etc.
- The formats (standalone and/or ISOBMFF) may store one or more IA Sequences
- Readers shall support sequences.

sunghee-hwang: Why do we need the next non-redundant Magic Code OBU?
If we signals the end of IA sequence(s), don't you think that we need to use different type of OBU or out of band signal.

cconcolato: Redundant OBUs (whatever their type) should not matter except if they are the first one encountered by the reader. Otherwise, it means adding any redundant OBUs anywhere would change the semantics and processing of the stream.

sunghee-hwang: Oops~. I misunderstood that the next non redundant Magic Code OBU was a part of the current IA sequence to signal its end.

sunghee-hwang: If ISOBMFF stores multiple IA sequences, do you intend to support multiple sample entries?

tdaede:
We discussed this and ultimately we can define that IA sequences can be started with a new non redundant magic code OBU, and that it starts a new IA Sequence. We will not make specific decoding guarantees about multiple sequential sequences.

tdaede (approved):

cconcolato (approved):

PR #427, opened 2023-05-22
Fix #402, loudness for Mono

### LINKED ISSUE #402: Allow Scalable Channel Layout with the highest layer mono
[Section 3.8](https://aomediacodec.github.io/iamf/#obu-mixpresentation) has a few phrases that seem to prevent using an Audio Element with only a mono layer.

1. Each sub-mix shall include [loudness_layout](https://aomediacodec.github.io/iamf/#loudness_layout) to identify [Loudspeaker configuration for Sound System A (0+2+0)](https://www.itu.int/dms_pubrec/itu-r/rec/bs/R-REC-BS.2051-3-202205-I!!PDF-E.pdf#) (i.e. Stereo). In other words, each sub-mix shall include loudness_info() for Stereo.
2. "In other words, rendering from an audio element with the highest layout to the [loudness_layout](https://aomediacodec.github.io/iamf/#loudness_layout) should not require an upmix."
3. The set of [loudness_layout](https://aomediacodec.github.io/iamf/#loudness_layout)s shall include all of [loudspeaker_layout](https://aomediacodec.github.io/iamf/#loudspeaker_layout)s specified in the [channel_audio_layer_config()](https://aomediacodec.github.io/iamf/#channel_audio_layer_config)s of Audio Element OBU for the [audio_element_id](https://aomediacodec.github.io/iamf/#audio_element_id).

By rule 1 we must include a stereo loudness for an Audio Element with only a single mono layer.. By Rule 2 it is not allowed to upmix mono to stereo in order to measured the loudness required in rule 1. Rules 1 and 2 seem to conflict and prevent an Audio Element with only a single mono layer.

By rule 3 we must also include loudness measured for mono. Rule 3 it is not exactly clear what the values to use in `class layout()` for the mono layer (even when it a mono layer with other layers on top.)

`layout()` does not have a `sound_system` for mono. So is the loudness for mono taken to mean a `LOUDSPEAKER_SP_LABEL` with a single loudspeaker? Does the `sp_label` of this single loudspeaker matter?

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/427.html" title="Last updated on May 23, 2023, 9:13 PM UTC (79f54ad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/427/d1b7b59...79f54ad.html" title="Last updated on May 23, 2023, 9:13 PM UTC (79f54ad)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1018,14 +1018,13 @@ The layout specified in [=loudness_layout=] should not be higher than the highes
 - [=loudness_layout=] for zero-order Ambisonics or Mono should not be higher than Stereo. Zero-order Ambisonics or Mono may be rendered to Stereo.
 
 If one sub-mix of Mix Presentation OBU includes only one single scalable channel audio, then it complies with as follows:
-- [=num_layouts=] shall be greater than or equal to [=num_layers=] specified in [=scalable_channel_layout_config()=] of Audio Element OBU for the [=audio_element_id=].
-- If the Audio Element is not Mono, then the set of [=loudness_layout=]s shall include all of [=loudspeaker_layout=]s specified in the [=channel_audio_layer_config()=]s of Audio Element OBU for the [=audio_element_id=].
+- [=num_layouts=] shall be greater than or equal to [=num_layers=] specified in [=scalable_channel_layout_config()=] of Audio Element OBU for the [=audio_element_id=] except the following cases:
 
 The highest [=loudness_layout=] specified in one sub-mix except zero-order Ambisonics or Mono is the layout which was used for authoring the sub-mix.
 - The highest [=loudness_layout=] for zero-order Ambisonics or Mono is Stereo.
  
 Each sub-mix shall include [=loudness_layout=] to identify [=Loudspeaker configuration for Sound System A (0+2+0)=] (i.e. Stereo). In other words, each sub-mix shall include loudness_info() for Stereo. 
-- If the mixed audio element by each sub-mix is Mono, then its [=loudness=] should be measured on Stereo which is generated from the Mono by the equations, L = 0.707 x Mono and R = 0.707 x Mono. 
+- If the mixed audio element by each sub-mix is Mono, then its [=loudness=] for [=loudness_layout=] = Stereo should be measured on the Stereo generated from the Mono by the equations, L = 0.707 x Mono and R = 0.707 x Mono. 
 
 ### Mix Presentation Annotations Syntax and Semantics ### {#obu-mixpresentation-annotation}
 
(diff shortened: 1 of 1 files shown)


sunghee-hwang (commented):
Added Mono in Sound System and updated descriptions related to Mono.

index.bs
```diff
@@ -1012,17 +1012,19 @@ class mix_presentation_obu() {
...
-- The set of [=loudness_layout=]s shall include all of [=loudspeaker_layout=]s specified in the [=channel_audio_layer_config()=]s of Audio Element OBU for the [=audio_element_id=]. 
-
-The highest [=loudness_layout=] specified in one sub-mix is the layout which was used for authoring the sub-mix.
+- [=num_layouts=] shall be greater than or equal to [=num_layers=] specified in [=scalable_channel_layout_config()=] of Audio Element OBU for the [=audio_element_id=] except the following cases:
 
+The highest [=loudness_layout=] specified in one sub-mix except zero-order Ambisonics or Mono is the layout which was used for authoring the sub-mix.
+- The highest [=loudness_layout=] for zero-order Ambisonics or Mono is Stereo.
+ 
 Each sub-mix shall include [=loudness_layout=] to identify [=Loudspeaker configuration for Sound System A (0+2+0)=] (i.e. Stereo). In other words, each sub-mix shall include loudness_info() for Stereo. 
+- If the mixed audio element by each sub-mix is Mono, then its [=loudness=] for [=loudness_layout=] = Stereo should be measured on the Stereo generated from the Mono by the equations, L = 0.707 x Mono and R = 0.707 x Mono.
```
shawn-singh, line 1027: I think this requirement (weighting mono loudness by 0.707 to indicate measured loudness) may be incorrect.  My understanding is that we would want a "dual mono" audio signal to correctly measure +3 dB louder than the corresponding mono, and then the playback system is expected to compensate accordingly depending on whether it plays back mono as true mono or dual mono.  It would be great to confirm with someone like Phil or other loudness experts, I can reach out to some loudness experts if needed.

sunghee-hwang: @shawn-singh 
Thank you for your opinion. Can you check the behavior of libear? What I heard is that libear renders from Mono to Stereo (L: -3dB x Mono, R: -3dB x Mono).

jwcullen (approved):
These updates resolve my concerns in the original issue; the rules now allow Audio Elements with only either a mono layer (either as the only layer or the first layer). 

Before merging we should open a smaller-scoped issue to discuss the concern brought up by @shawn-Singh about loudness measurement weighting for mono.

yeroro (approved):

index.bs
shawn-singh: If libear is applying -3 dB when rendering mono into dual mono, I think that aligns with what I'm saying earlier - the playback system is expected to compensate if they play back mono as dual mono.   But to represent the actual loudness measurement, I think we do not need to apply a 0.707 adjustment.

phastphill: The syntax change makes sense to me, in that it specifies that the upmix from mono to stereo would always be energy-preserving, which is typical.

I do think that, for a mono signal M, monoLoudness(M) ==  stereoLoudness(.707M, .707M), so perhaps this is unnecessary?

PR #428, opened 2023-05-22
Fix #403, ISOBMFF Improvement

### LINKED ISSUE #403: ISOMBFF Encapsulation
1. Section 6.2.1 https://aomediacodec.github.io/iamf/#isobmff-singletrack
It should introduce the fact that even though an IA sequence can theoretically group audio data coded with different codecs, potentially with different timing properties, which would require multiple tracks, this version of the specification only supports storing an IA Sequence as a single track thanks to the restrictions of the selected profiles. Also consider merging that section with the sentence above that is not specific enough (does not mention single track).

2. Section 6.2.2. Encapsulation Scheme
Instead of defining how IAMF is encapsulated, in terms of algorithms, the section should be written in terms the end result and of shall statements describing the validity of a IAMF file. This will help writing validators, like [ComplianceWarden](https://github.com/gpac/ComplianceWarden/)

I suggest rephrasing:
> During encapsulation process, OBUs of IA sequences are encapsulated into [[ISOBMFF]](https://aomediacodec.github.io/iamf/#biblio-isobmff) as follows:

to:
> The result of encapsulating an IA Sequence into an [[ISOBMFF]](https://aomediacodec.github.io/iamf/#biblio-isobmff) file is as follows:

3. Sample Format 
The text about "IA Sample" in 6.2.2 should be moved to section 6.2.5. This would group all the spec text around sample constraints in a single place.
> Temporal Delimiter OBU: shall be discarded if present.

Instead of talking about the process of a packager, describe the end result, rephrase to say: 
> Sample data shall not contain Temporal Delimiter OBUs.

> Temporal Unit that has no OBUs is discarded

It does not seem to be possible to have an empty Temporal Unit given the definition.

> Temporal units shall be placed in consecutive IA Samples in order.

This sentence is ambiguous: 
- does it talk about "byte order"? It shouldn't because editing tools have the liberty to place samples anywhere in the file as long as offsets are properly documented. 
- does it talk about "time order". Given that each Temporal Unit has a start time, and that each Sample is formed by a Temporal Unit, with DTS being the TU start time, the time order follows. 
I suggest deleting this sentence.

> One IA Sample data shall be one or more consecutive temporal units (after removing Temporal Delimiter OBUs if present) for every IA Sample to be sync sample.

I don't understand this sentence:
- Why is the spec allowing more than one temporal unit per sample? This prevents precise seeking. What is the limit? Can I put all TU into 1 single sample? 
- the part "after removing Temporal Delimiter OBUs if present" should be removed (covered above with a clear "shall")
- the part "for every IA Sample to be sync sample" is not clear. What does "for" mean? It could simply say in another sentence: "Each IA Sample shall be marked as a sync sample". But is it what we want? If one day we want to use a audio codec that uses non-sync samples, it would be a problem. We could say something like: 
> If all the Audio Frame OBUs in the IA Sample contain coded data that would be sync sample if they were stored in their own non-IAMF track, then the IA Sample shall be marked as sync sample.

or :
> If all the Audio Frame OBUs in the IA Sample contain coded data that can be decoded only with the information contained in the the sample description entry, then the IA Sample shall be marked as sync sample.

> In other words, every IA Sample consists of all the data OBUs that are associated with a specific, distinct time instant.

What is this sentence trying to say? I suggest removing it.

> If [num_samples_to_trim_at_start](https://aomediacodec.github.io/iamf/#num_samples_to_trim_at_start) < [num_samples_per_frame](https://aomediacodec.github.io/iamf/#num_samples_per_frame), every IA Sample except the last IA Sample shall have the same number (N > 0) of temporal units.

It's not clear how the trimming concept is related to the concept of Temporal Unit. If they are, the concept should be defined in a non-ISOBMFF-specific section of the spec.

> In other words, every IA Sample except the last IA Sample consists of N consecutive Audio Frame OBUs per substream and one or more Parameter Block OBUs per parameter (if present), the total duration of the one or more Parameter Block OBUs is equal to the total duration of the N consecutive Audio Frame OBUs.

First, an ISOBMFF track could contain multiple, consecutive IA Sequences in the same track. So, the concept of mapping the "last IA Sample" of a track to trimming is probably wrong. Rather the specification should define the duration of an IA sample. 
Then, it should be clarified that if trimming is used in one Audio Frame OBU of a sample for a given substream, all the other substreams have to have the same otherwise you cannot easily map to edit list or durations. 

I suggest:
> The decode duration of an IA Sample SHALL equal the duration of the underlying temporal unit, i.e. the sum of the decode durations of all the Audio Frame OBUs of any single substream. 
> NOTE: Per the restriction of the profiles carried in an IA track, all Audio Frame OBUs in an IA Sample have the same duration and have the same trimming information. 
> If Audio Frame OBUs in the sample contain trimming information, the corresponding audio samples SHALL be removed from presentation using edit list information.
> NOTE: In typical case, when a track contains a single IA Sequence, trimming can only happen at the beginning or end of the IA sequence and therefore at the beginning or end of the track and the edit list can describe the start and end trimming with a single edit entry. Track storing consecutive IA Sequences may need multiple edits in the edit list.

> In every IA Sample, the relative offset of each substream or each parameter (if present) is identified by its associated [relative_offset](https://aomediacodec.github.io/iamf/#relative_offset) field of the sync_obu() in [IA_Descriptor](https://aomediacodec.github.io/iamf/#ia_descriptor).

IIUC, this is bad because seeking into the file (using usual ISOBMFF procedure: edit list, decode time, presentation time) will lead to the wrong timing.

4.[IA Descriptor](https://aomediacodec.github.io/iamf/#iadescriptors-section)
First there is a mismatch between the section name (plural) and the class name (singular)
What is the need to define `iamd` compared to having a simple array of bytes in `iamf`, as is done with `configOBUs` for AV1 (or other video codecs). You could simply say that `configOBUs` shall contain the following OBUs in order.

> All descriptor OBUs in the IA sequences are then discarded from the IA sequence.

Instead of 'discarded', the spec should talk about the results:
> Sample data SHALL not contain Descriptor OBUs.

5. Sample Entry
- sample rate: Given the restrictions in the profile, it seems that the actual data in the track has a single sample rate, why not expose it in the sample entry?
- Similarly, why not expose the channel count, when it makes sense? 
- There should be a note that says that multiple sample entries may be used in a track, for example when the track is the concatenation of multiple tracks or multiple IA Sequence and some samples have different configOBUs values.

6. Sample Group
It is not clear what the following means:
> The [roll_distance](https://aomediacodec.github.io/iamf/#roll_distance) value in the Codec Config OBU is copied to [AudioRollRecoveryEntry](https://www.iso.org/standard/68960.html#) with the [grouping_type](https://www.iso.org/standard/68960.html#) = [roll](https://www.iso.org/standard/68960.html#).

Does it mean that a Sample Group of type `roll` shall be present? BTW, the Sample Group used to indicate Pre-Roll is `prol` not `roll` (see CMAF for example). Maybe you want to say (and update Figure 6):
> In an IA Track, every sample shall be associated with a sample group of type `prol`. The `roll_distance` value SHALL equal the value of the `roll_distance` field in the Codec Config OBU stored in the `configOBUs` array in the sample entry.

7. Track Header and Movie Header

ISOBMFF defines how to set the `mvhd` based on the information in the tracks of the file. I suggest removing mentions of the `mvhd`.
Since trimming is covered by the edit list, and since ISOBMFF covers how to set (or compute) the track duration, I suggest removing any mentions about track duration.

8. Decapsulation process

The text should not talk about the "i-th" sample entry. Sample entries can be reordered for various reasons. It should take the sample in presentation order and when a sample description index is changed, then the content of the `configOBUs` is inserted in the sequence.

Since trimming information at the container level (edit list) reflects the content of the OBUs in the track, I don't see the need for section 7.1.1. Also, I don't understand that section (PTS, PTS1, PTS2 ...).

I suggest merging this section as an informative section in section 7 and changing the title of section 6 to "ISOBMFF IAMF Track Format"

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/428.html" title="Last updated on May 26, 2023, 6:49 PM UTC (49edd09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/428/d1b7b59...49edd09.html" title="Last updated on May 26, 2023, 6:49 PM UTC (49edd09)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1926,7 +1926,7 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
 
 ```
 class IASampleEntry extends AudioSampleEntry('iamf') {
-  unsigned int 8[] configOBUs;
+  unsigned int (8) configOBUs[];
 }
 ```
 
--- a/index.bs
+++ b/index.bs
@@ -1884,7 +1884,7 @@ A file conformant to this specification satisfies the following:
 Parsers shall support the structures required by the <code>'iso6'</code> brand and MAY support structures required by further ISOBMFF structural brands.
 
 
-## ISOBMFF IAMF Encapsulation with single track ## {#isobmff-singletrack}
+## ISOBMFF IAMF Encapsulation ## {#isobmff-singletrack}
 
 This section describes the basic data structures used to signal encapsulation of IA sequence in [[!ISOBMFF]] containers.
 
--- a/index.bs
+++ b/index.bs
@@ -2002,9 +2002,9 @@ For example, for this version of the specification
 	iamf.0000.0100.mp4a.40.2
 ```
 
-# ISOBMFF IAMF Decapsulation # {#isobmff-decapsulation}
+## ISOBMFF IAMF Decapsulation (Informative) ## {#isobmff-decapsulation}
 
-## ISOBMFF IAMF Decapsulation with single track ## {#isobmff-decapsulation-singletrack}
+### ISOBMFF IAMF Decapsulation with single track ### {#isobmff-decapsulation-singletrack}
 
 This section provides a guideline for IAMF parser to reconstruct IA sequences from IAMF file.
 
--- a/index.bs
+++ b/index.bs
@@ -1896,11 +1896,11 @@ Even though an IA sequence can theoretically group audio data coded with differe
 
 The result of encapsulating an IA Sequence into an [[!ISOBMFF]] file is as follows:
 
-- If there is audio samples to be trimmed at the start or at the end, then 'edts' and 'elst' boxes shall be present to reflect the trimming status.
+- If there are audio samples to be trimmed at the start or at the end, then 'edts' and 'elst' boxes shall be present to reflect the trimming status.
 - Sample Entry
 	- The first [=IASampleEntry=] shall include the first Descriptor OBUs and followed by Sync OBU of the IA sequence. 
 	
-NOTE: Multiple sample entries may be used in a track, for example when the track is the concatenation of multiple tracks or multiple IA Sequences and some samples have different configOBUs values.
+NOTE: Multiple sample entries may be used in a track, for example when the track is the concatenation of multiple tracks or multiple IA Sequences and some IA samples have different configOBUs values.
 
 - Decoding Time to IA Sample
 	- 'stts' or 'trun' box shall indicate the number of audio samples which [=IA Sample=] includes (i.e. the duration of IA Sample). 
@@ -1951,7 +1951,7 @@ For tracks using the [=IASampleEntry=], an <dfn noexport>IA Sample</dfn> has the
 		- In other words, [=relative_offset=](s) in sync_obu() is (are) applicable to every IA Sample in its associated [=IASampleEntry=].
 - The decode duration of an IA Sample SHALL equal the duration of the underlying temporal unit, i.e. the decode durations of the Audio Frame OBU.
 
-NOTE: Per the restriction of the profiles carried in an IA track, all Audio Frame OBUs in an IA Sample have the same duration and have the same trimming information. If Audio Frame OBUs in the sample contain trimming information, the corresponding audio samples SHALL be removed from presentation using edit list information.
+NOTE: Per the restriction of the profiles carried in an IA track, all Audio Frame OBUs in an IA Sample have the same duration and have the same trimming information. If Audio Frame OBUs in the IA sample contain trimming information, the corresponding audio samples SHALL be removed from presentation using edit list information.
 
 NOTE: In typical case, when a track contains a single IA Sequence, trimming can only happen at the beginning or end of the IA sequence and therefore at the beginning or end of the track and the edit list can describe the start and end trimming with a single edit entry. Track storing consecutive IA Sequences may need multiple edits in the edit list.
 
--- a/index.bs
+++ b/index.bs
@@ -1898,7 +1898,7 @@ The result of encapsulating an IA Sequence into an [[!ISOBMFF]] file is as follo
 
 - If there are audio samples to be trimmed at the start or at the end, then 'edts' and 'elst' boxes shall be present to reflect the trimming status.
 - Sample Entry
-	- The first [=IASampleEntry=] shall include the first Descriptor OBUs and followed by Sync OBU of the IA sequence. 
+	- For an IA Sample, the required Descriptor OBUs and Sync OBUs required for processing the samples shall be in the [=configOBUs=] of the associated sample entry. 
 	
 NOTE: Multiple sample entries may be used in a track, for example when the track is the concatenation of multiple tracks or multiple IA Sequences and some IA samples have different configOBUs values.
 
@@ -1926,7 +1926,7 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
 
 ```
 class IASampleEntry extends AudioSampleEntry('iamf') {
-  configOBUs;
+  unsigned int 8[] configOBUs;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1941,63 +1936,24 @@ None of AudioSampleEntry's optional boxes shall be present.
...
-- If [=num_samples_to_trim_at_start=] < [=num_samples_per_frame=], every IA Sample except the last IA Sample shall have the same number (N > 0) of temporal units.
-	- In other words, every IA Sample except the last IA Sample consists of N consecutive Audio Frame OBUs per substream and one or more Parameter Block OBUs per parameter (if present), the total duration of the one or more Parameter Block OBUs is equal to the total duration of the N consecutive Audio Frame OBUs. 
-	- In every IA Sample, the relative offset of each substream or each parameter (if present) is identified by its associated [=relative_offset=] field of the sync_obu() in [=IA_Descriptor=].
+For tracks using the [=IASampleEntry=], an <dfn noexport>IA Sample</dfn> has the following constraints:
+- One IA Sample data shall be one Temporal Unit and shall not contain Temporal Delimiter OBU.
+- In every IA Sample, the relative offset of each Substream or each Parameter (if present) is identified by its associated [=relative_offset=] field of the sync_obu().
 		- In other words, [=relative_offset=](s) in sync_obu() is (are) applicable to every IA Sample in its associated [=IASampleEntry=].
```
cconcolato, line 1951: I would delete these 2 sentences because the timing of a Sample in ISOBMFF is fully defined by its duration

index.bs
```diff
@@ -1884,34 +1884,29 @@ A file conformant to this specification satisfies the following:
...
-	- Take the number of trimmed samples from the IA sequence and converting it to the sample rate used by the 'elst' boxes.
-	- Reflect the converted numbers to 'mvhd' to indicate the duration after trimming.
-	- If there is audio samples to be trimmed at the start or at the end, then 'edts' and 'elst' boxes shall be present to reflect the the number of audio samples to be trimmed at the start of the IA sequence (i.e [=media_time=]) and [=edit_duration=].
+- If there are audio samples to be trimmed at the start or at the end, then 'edts' and 'elst' boxes shall be present to reflect the trimming status.
 - Sample Entry
-	- The ith set of descriptor OBUs which is not a redundant copy of the previous descriptor OBUs (i.e. the first descriptor OBUs is regarded as non-redundant copy regardless of its [=obu_redundant_copy=]), specified in [[#standalone-descriptor-obus]], and the Sync OBU that immediately follows them are moved to the ith [=IASampleEntry=]. All descriptor OBUs in the IA sequences are then discarded from the IA sequence. 
+	- The first [=IASampleEntry=] shall include the first Descriptor OBUs and followed by Sync OBU of the IA sequence.
```
cconcolato, line 1901: We should say that for an IA Sample, the required Descriptor OBUs and Sync OBUs required for processing the samples shall be in the configOBUs of the associated sample entry.

index.bs
```diff
@@ -1925,13 +1920,13 @@ During encapsulation process, OBUs of IA sequences are encapsulated into [[!ISOB
...
 
-<dfn noexport>IASampleEntry</dfn> identifies that the track contains [=IA Samples=], and contains one IA_Descriptor.
+<dfn noexport>IASampleEntry</dfn> identifies that the track contains [=IA Samples=], and contains configOBUs.
 
 <b>Syntax</b>
 
 ```
 class IASampleEntry extends AudioSampleEntry('iamf') {
-  IA_Descriptor descriptor;
+  configOBUs;
```
cconcolato, line 1929: The type should be `byte[]`

cconcolato (approved):

index.bs
```diff
@@ -1926,7 +1926,7 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
 
 ```
 class IASampleEntry extends AudioSampleEntry('iamf') {
-  configOBUs;
+  unsigned int 8[] configOBUs;
```
cconcolato, line 1929: AV1-ISOBMFF uses the following syntax
```suggestion
  unsigned int (8) configOBUs[];
```

PR #431, opened 2023-05-23
Fix #421, add default value for demixing and recon gain

### LINKED ISSUE #421: Consider DemixingParamDefinition() having default values for demixing parameters
It is better to add default_dmixp_mode and default_w value so that we can reduce its bandwidth in case of static demixing.

sunghee-hwang:
- Updated Audio Element OBU syntax and semantic section
 - Explicitly mentioned there may be no parameter block obus for demixing or recon gain.
 - Default values are used when no parameter block obu presents.
 - Added syntaxes for default demixing parameter data

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/431.html" title="Last updated on Jun 2, 2023, 2:06 PM UTC (330e1ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/431/795601b...330e1ac.html" title="Last updated on Jun 2, 2023, 2:06 PM UTC (330e1ac)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -649,6 +649,9 @@ class audio_element_obu() {
 
 ```
 class DemixingParamDefinition() extends ParamDefinition() {
+  default_demixing_info_parameter_data();
+  unsigned int (4) default_w;
+  unsigned int (4) reserved;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jwcullen (approved):
To keep this moving forward quickly I put in a commit to change the wording about default recon gain that I think reads a little clearer. The meaning should be the same.

I approve that this PR makes the technical changes agreed on in the recent meeting. If others have comments about editorial issues they can make comments - otherwise this should be merged in and editorial issues can be fixed later.

yeroro (approved):

PR #435, opened 2023-05-24
Fix #410, clarified usage of demixing and recon gain

### LINKED ISSUE #410: Need to clarify if it is possible scalable channel audio not having demixing info and/or not having recon_gain info.
It is unclear when the parameters are present and when they are not present. And, it is unclear what is decoders' behaviours when it is not present
- [num_layers](https://aomediacodec.github.io/iamf/#num_layers) shall be set to 1 or up to 6 for Channel-based audio element (i.e. scalable channel audio).
  - In this case, [demixing_info_parameter_data()](https://aomediacodec.github.io/iamf/#demixing_info_parameter_data) and [recon_gain_info_parameter_data()](https://aomediacodec.github.io/iamf/#recon_gain_info_parameter_data) may be present in the IA sequence.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/435.html" title="Last updated on May 24, 2023, 2:14 AM UTC (e765a96)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/435/795601b...e765a96.html" title="Last updated on May 24, 2023, 2:14 AM UTC (e765a96)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -690,8 +690,17 @@ A ChannelGroup is defined in [[#iamfgeneration]]. The order of the substreams in
 
 
 <dfn noexport>num_parameters</dfn> specifies the number of parameters that are used by the algorithms specified in this audio element.
+- When [=audio_element_type=] = 0, this field shall be set to 0, 1 or 2.
+- When [=audio_element_type=] = 1, this field shall be set to 0.
 
 <dfn noexport>param_definition_type</dfn> specifies the type of the parameter definition. All parameter definition types described in this version of the specification are listed in the table below, along with their associated parameter definitions.
+- The type PARAMETER_DEFINITION_MIX_GAIN shall not be present in Audio Element OBU.
+- The type shall not be duplicated in one Audio Element OBU.
+- When [=codec_id=] = 'fLaC' or 'ipcm', the type PARAMETER_DEFINITION_RECON_GAIN shall not be present.
+- When [=num_layers=] > 1, the type PARAMETER_DEFINITION_RECON_GAIN shall be present.
+- When the highest [=loudspeaker_layout=] of (non-)scalable channel audio is less than or equal to 3.1.2ch, the type PARAMETER_DEFINITION_DEMIXING shall not be present.
+- When the highest [=loudspeaker_layout=] of scalable channel audio ([=num_layers=] > 1) is greater than 3.1.2ch, the types of both PARAMETER_DEFINITION_DEMIXING and both PARAMETER_DEFINITION_RECON_GAIN shall be present.
+- When [=num_layers=] = 1 and [=loudspeaker_layout=] is greater than 3.1.2ch, the type PARAMETER_DEFINITION_DEMIXING may be present.
 
 <table class = "def">
 <tr>


jwcullen:
What is supposed to happen for a bitstream with the following properties?

- 1 scene based Audio Element
- A Mix Gain OBU with a stereo and 7.1.4 layout.
- The listener wants to listen to 3.1.2.

Section [8.3.2.2](https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/435/795601b...e765a96.html#processing-mixpresentation-rendering-a2l) tells me that 

`The IA renderer used is selected according to the following rules:`
...
`Else, use implementation-specific renderer.`
`If there is no implementation-specific Ambisonics renderer, use the EAR HOA renderer to render to the next highest`

The above rules mean that the  scene based Audio element should be rendered to 7.1.4 (this is OK). But the next sentence:
`layout compared to the playback layout, and then downmix using implementation-specific renderer.`

This rule means that now the 7.1.4 should be down-mixed to 3.1.2 to support to user's request. But this PR prevents scene based audio elements from having any parameters (e.g. demixing). 

I think it may make sense to allow an "implementation-specific" rendering to use the demixing parameters in this case if they are provided. Maybe it would be useful to allow demixing parameters to support this type of use case? Or do I misunderstand something about this.

sunghee-hwang:
> What is supposed to happen for a bitstream with the following properties?
> 
> * 1 scene based Audio Element
> * A Mix Gain OBU with a stereo and 7.1.4 layout.
> * The listener wants to listen to 3.1.2.
> 
> Section [8.3.2.2](https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/435/795601b...e765a96.html#processing-mixpresentation-rendering-a2l) tells me that
> 
> `The IA renderer used is selected according to the following rules:` ... `Else, use implementation-specific renderer.` `If there is no implementation-specific Ambisonics renderer, use the EAR HOA renderer to render to the next highest`
> 
> The above rules mean that the scene based Audio element should be rendered to 7.1.4 (this is OK). But the next sentence: `layout compared to the playback layout, and then downmix using implementation-specific renderer.`
> 
> This rule means that now the 7.1.4 should be down-mixed to 3.1.2 to support to user's request. But this PR prevents scene based audio elements from having any parameters (e.g. demixing).
> 
> I think it may make sense to allow an "implementation-specific" rendering to use the demixing parameters in this case if they are provided. Maybe it would be useful to allow demixing parameters to support this type of use case? Or maybe this is not a reasonable way to use demixing parameters?

As libear does not support 3.1.2ch output, our spec has defined static down-mix matrics to address it.
I think that the matrics should be used for the case.

Demixing info has not been considered for scene-based. If you feel that it would be useful scenario Ambisonics with demixing info, how about discussing it at the stage for V2?

index.bs
```diff
@@ -690,8 +690,17 @@ A ChannelGroup is defined in [[#iamfgeneration]]. The order of the substreams in
...
+- When [=audio_element_type=] = 1, this field shall be set to 0.
 
 <dfn noexport>param_definition_type</dfn> specifies the type of the parameter definition. All parameter definition types described in this version of the specification are listed in the table below, along with their associated parameter definitions.
+- The type PARAMETER_DEFINITION_MIX_GAIN shall not be present in Audio Element OBU.
+- The type shall not be duplicated in one Audio Element OBU.
+- When [=codec_id=] = 'fLaC' or 'ipcm', the type PARAMETER_DEFINITION_RECON_GAIN shall not be present.
+- When [=num_layers=] > 1, the type PARAMETER_DEFINITION_RECON_GAIN shall be present.
+- When the highest [=loudspeaker_layout=] of (non-)scalable channel audio is less than or equal to 3.1.2ch, the type PARAMETER_DEFINITION_DEMIXING shall not be present.
+- When the highest [=loudspeaker_layout=] of scalable channel audio ([=num_layers=] > 1) is greater than 3.1.2ch, the types of both PARAMETER_DEFINITION_DEMIXING and both PARAMETER_DEFINITION_RECON_GAIN shall be present.
+- When [=num_layers=] = 1 and [=loudspeaker_layout=] is greater than 3.1.2ch, the type PARAMETER_DEFINITION_DEMIXING may be present.
```
yeroro, line 703: I am curious: when will the demixing info be used when `num_layers = 1`?

sunghee-hwang: For example, 7.1.4ch non scalable audio may be delivered with a demixing info. Then devices should do dynamic down-mix from 7.1.4ch to its playback layout for example 3.1.2ch or 5.1.2ch by using the demixing info.
When demixing info does not present, as you know the down-mixing should be based on libear.

sunghee-hwang:
> This rule means that now the 7.1.4 should be down-mixed to 3.1.2 to support to user's request. But this PR prevents scene based audio elements from having any parameters (e.g. demixing).

What this PR prevent is any parameters in Audio Element OBU when it is scene-based. But Parameters can be still defined in Mix Presentation OBUs.

yeroro (approved):

PR #438, opened 2023-05-25
Fix #400, Audio Frame OBU improvement

### LINKED ISSUE #400: Audio Frame OBU improvement suggestions
I find the term "Audio Frame OBU" a bit hard to understand:
- It is defined as "the raw coded audio frame for an audio substream." 
- It is represented by different types of OBUs `audio_frame_obu_with_no_id` and `audio_frame_obu`
- The `audio_frame_obu_with_no_id` actually starts with an id `audio_substream_id`
- the term Audio OBU (https://aomediacodec.github.io/iamf/#audio-obu) is too close to Audio Frame OBU

I would suggest:
- removing "raw" from the definition
- adding to the definition that "it is represented by different types of OBUs"
- renaming `audio_frame_obu` to `audio_frame_obu_explicit_id` and `audio_frame_obu_with_no_id` to `audio_frame_obu_implicit_id`
- renaming the top level structure from `audio_open_bitstream_unit` to `ia_ open_bitstream_unit`

sunghee-hwang:


jwcullen (changes requested):
I think the terminology of implicit and explicit are swapped in several places in this commit.

For example the commit has this wording:
```
class audio_frame_obu_implicit_id() {
  leb128() audio_substream_id;
  audio_frame_obu_explicit_id(audio_substream_id);
}
```
But in my opinion this class has the ID explicitly listed and would better be called `audio_frame_obu_explicit_id` with an inner class of `audio_frame_obu_implicit_id`.

The commit also has this wording:
```
class audio_frame_obu_explicit_id() {
  unsigned int (8*coded_frame_size) audio_frame();
}
```
In my opinion the ID is implicit to this class (from the OBU header) or from being nested within the previous class. It would better be called `audio_frame_obu_implicit_id`.

sunghee-hwang:
> I think the terminology of implicit and explicit are swapped in several places in this commit.
> 
> For example the commit has this wording:
> 
> ```
> class audio_frame_obu_implicit_id() {
>   leb128() audio_substream_id;
>   audio_frame_obu_explicit_id(audio_substream_id);
> }
> ```
> 
> But in my opinion this class has the ID explicitly listed and would better be called `audio_frame_obu_explicit_id` with an inner class of `audio_frame_obu_implicit_id`.
> 
> The commit also has this wording:
> 
> ```
> class audio_frame_obu_explicit_id() {
>   unsigned int (8*coded_frame_size) audio_frame();
> }
> ```
> 
> In my opinion the ID is implicit to this class (from the OBU header) or from being nested within the previous class. It would better be called `audio_frame_obu_implicit_id`.

Thanks for your opinion.
Please see the suggestion on issue #400.
It is opposite. Let's talk about this at the call.

tdaede:
I suggest either swapping the implicit and explicit naming, or coming up with a different naming entirely (e.g. audio_frame_obu_extended, or provide an arg to audio_frame_obu)

tdaede:
@jwcullen will propose an alternate text

index.bs
```diff
@@ -613,7 +613,7 @@ class codec_config() {
 
 <dfn noexport>codec_id</dfn> indicates a ‘four-character code’ (4CC) to identify the codec used to generate the audio substreams. It shall be 'Opus' for IAMF-OPUS, 'mp4a' for IAMF-AAC-LC, 'fLaC' for IAMF-FLAC and 'ipcm' for IAMF-LPCM.
 
-<dfn noexport>num_samples_per_frame</dfn> indicates the frame length, in samples, of the raw coded audio provided in by audio_frame_obu(). It shall not be set to zero.
```
tdaede, line 616: we should make sure to not drop "raw"

Issue #447, opened 2023-05-26
Consider increasing size of magic code


tdaede:
To make missdetections less likely.

sunghee-hwang:
@tdaede, 
What would be the best candidates?
- iamfiamf (64bits)
- aomiamf (56bits)
- iaiamf (48bits)

tdaede:
dupe of #441

Issue #441, opened 2023-05-25
Searching for magic code is not a guarantee


tdaede:
We currently say:

> NOTE: When IA OBUs are delivered over a protocol that does not provide explicit IA Sequence boundaries, a parser may locate the IA Sequence start by searching for the code 'iamf' preceded by specific OBU header values, e.g. assuming [=obu_extension_flag=] is set to 0 and because [=obu_trimming_status_flag=] is set to 0 for a Magic Code OBU, the OBU header can be 0xF806 or 0xFC06

It would be good to expand on this a bit that finding that byte value is not a guarantee that the start of the sequence is found (it could occur within other OBUs)

sunghee-hwang:
We may need to add this kind of logic got get sync pattern to Informative Annex.

The logic to get sync pattern would be like this;
- Step1: Check if the first 7bits is 1111100 or 1111110
- Step2: Check if the next 1 bit is 0(no extension) or 1 (extension)
- Step3-1: If there is no extension, then check if # of bytes of `obu_size` field and followed by `ia_code`.
- Step3-2: If there is extension, check if # of bytes of `obu_size` field, followed by # of bytes of extension header indicated by `extension_header_size` field and followed by `ia_code`.

So, the size of sync pattern would be regarded as at least 38 bits = 6bits (in step1) + 32bits (in step3). (There is a condition in step2, so actual bits would be more than this)
Its probability to occur within other OBUs in accident is about 2.7 x 10 ^(-11).
If we assume IAMF bitrate of 768kbps, it may occur one time per 3.5 x 10^5 second (one per 97.6 hours = about 4 days). Even broadcasting scenario, we have upper-layer signal which its payload is IAMF. So, actually probability in real world is much lower than this.

In addition, if we still feel uncomfortable, then we may add one more logic like this.
- Step4: Check if Magic Code OBU and followed by 7 bits = 0000000 or 0000010 (Codec Config OBU).
- (In this case, 2^6 = 64 times than the previous. I.e. one per 256 days)

If we need more,
-Step5: Check if Codec Config OBU and followed by 7 bits = 0000100 or 0000110. (one per 44.9 years)

tdaede:
That seems like enough for me.

Personally, I'd actually like to discourage seeking around for a magic code in general. Magic codes only occur at the beginning of a stream (so they aren't useful for random seeking), and we currently require streams to start with a magic code, so the utility of them is pretty low.

[...]

PR #453, opened 2023-05-30
Fix #445 and Fix #425, Magic Code Profiles

### LINKED ISSUE #425: Need to replace version 8bits and profile version 8 bits by profile name 8bits and compatible profile 8 bits.
To address the consensus on F2F meeting (May 25), version (8bits) and profile version(8bits) fields are replaced by profile_name(8bits) and compatible_profile fields(8bits)

### LINKED ISSUE #445: Backwards compatibility strategy
The spec should specify in detail how forwards and backwards compatibility works between versions.

The current plan that v2 streams will not be backwards compatible with v1 parsers. However, v2 parsers should be able to read v1 streams with little or no dedicated v2-specific logic.

sunghee-hwang:


sunghee-hwang:
It would be good if we could address issue #441 and #447.

yeroro (approved):

index.bs
```diff
@@ -554,36 +554,43 @@ class byte_alignment() {
...
 
-<dfn noexport>version</dfn> indicates the version of an IA sequence. It shall be set to 0 for this version of the specification. Implementations should treat IA sequences where the MSB four bits of the version number match that of a recognized specification as backwards compatible with that specification. That is, the version number can be split into "major" and "minor" version sub-fields, with changes to the minor sub-field (in the LSB four bits) signaling compatible changes. For example, an implementation of this specification should accept any stream with a version number of ’15’ or less, and should assume any stream with a version number ’16’ or
-greater is incompatible.
+<dfn noexport>profile_name</dfn> indicates that this IA sequence is able to be processed by IA decoders to support the profile indicated in this field. The IA decoders shall be able to parse all OBUs explicitly listed in that profile and can still encounter reserved OBUs that it should skip and it is acceptable to do so. This allows future versions of the specification to define new profiles that can be backwards compatible to old profiles.
```
cconcolato, line 583: Typo
```suggestion
<dfn noexport>profile_name</dfn> indicates that this IA sequence is able to be processed by IA decoders that support the profile indicated in this field. The IA decoders shall be able to parse all OBUs explicitly listed in that profile and can still encounter reserved OBUs that it should skip and it is acceptable to do so. This allows future versions of the specification to define new profiles that can be backwards compatible to old profiles.
```

index.bs
```diff
@@ -554,36 +554,43 @@ class byte_alignment() {
...
+- 0: Simple Profile
+- 1: Base Profile
+- 2~255: Reserved
 
-<dfn noexport>profile_version</dfn> indicates the profile of an IA sequence. The MSB four bits indicates the profile of an IA sequence. Implementations should treat IA sequences where the MSB four bits of the version number match that of a recognized profile as backwards compatible with that specification. That is, the version number can be split into "profile major" and "profile minor" version sub-fields, with changes to the minor sub-field (in the LSB four bits) signaling compatible changes with the profile major version. The semantic of this field is only valid when the MSB four bits of [=version=] = 0.
+<dfn noexport>profile_compatible</dfn> indicates the profile of specification which this IA sequence complies to.
```
cconcolato, line 588: What is "the profile"?
```suggestion
<dfn noexport>profile_compatible</dfn> indicates an additional profile of this specification to which this IA sequence complies. If a sequence only complies with the [=profile_name=] profile, this field shall be set to same [=profile_name=] value.  
```

index.bs
```diff
@@ -2020,25 +2028,20 @@ DASH and other applications require defined values for the 'Codecs' parameter sp
...
 
-<b>IAMF-specific-needs</b> shall be <b>V.PV</b> as follows:
-- <dfn noexport>V</dfn> is four digits and represents the [=version=] of IA sequence.
-	- The first two digits represents the major version within the range 0 to 15.
-	- The second two digits represents the minor version within the range 0 to 15.
-- <dfn noexport>PV</dfn> is four digits and represents the [=profile_version=] of IA sequence.
-	- The first <b>P</b> is two digits and represents the profile major version within the range 0 to 15.
-	- The second <b>V</b> is two digits and represents the profile minor version within the range 0 to 15.
+<b>IAMF-specific-needs</b> shall be <b>PC</b> as follows:
+- <dfn noexport>PC</dfn> is three digits within the range 0 to 255 and represents that IA decoder supports the profile of specification.
```
cconcolato, line 2032: The compatible profile should be listed in the codecs parameter.

sunghee-hwang: Let me share my assumption.
The Base profile is a super set of the Simple profile. The enhanced profile was also a super set of the Base profile.
If we define a new profile which is a super set of the most recent previous profile, we do not need to list all of compatible profile. 
What do you think?

tdaede: @cconcolato do you mean that you want two numbers, both PN and PC (000.000 for example) in the codec string?

tdaede: @sunghee-hwang will make a second PR for the other version number.

PR #503, opened 2023-06-22
Fix #493, Clarification on extension header

### LINKED ISSUE #493: Extensibility features for forwards compatibility
- Behavior of decoders should be specified for extension header. E.g. either the extension header should be discarded or the entire OBU should be discarded.
- Behavior for unknown codec_ids. Parsers should discard all mix presentations that reference elements in that codec_id.
- param_definition_type and audio_element_type cannot be expanded in a compatible way without using a new OBU.
- Behavior of loudspeaker_layout when a reserved value is used. That layer and all subsequent ones should be skipped.
- Adding non-mono/stereo substreams is not possible without breaking compatibility. The sentence "Mono or stereo coding shall be only allowed for the version of this specification." should be removed and enforced elsewhere.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/503.html" title="Last updated on Jun 27, 2023, 11:03 PM UTC (5bdeb37)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/503/f9167cc...5bdeb37.html" title="Last updated on Jun 27, 2023, 11:03 PM UTC (5bdeb37)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1591,7 +1591,7 @@ NOTE: The [=Temporal Delimiter OBU=] has an empty payload.
 
 ## Codec Specific ## {#codec-specific}
 
-This section defines codec specific information for Codec_Specific_Info and its coded [=Audio Substream=]. To generate one single coded [=Audio Substream=], mono or stereo coding SHALL be only allowed for this version of the specification.
+This section defines codec specific information for Codec_Specific_Info and its coded [=Audio Substream=]. To generate one single coded [=Audio Substream=], only mono or stereo coding SHALL be allowed for this version of the specification.
 
 - <dfn noexport>Codec_Specific_Info</dfn> is composed of [=codec_id=] and [=decoder_config()=]. 
 
--- a/index.bs
+++ b/index.bs
@@ -1441,7 +1441,7 @@ class mix_gain_parameter_data() {
 
 <b>Semantics</b>
 
-<dfn noexport>animation_type</dfn> specifies the type of animation applied to the parameter values.
+<dfn noexport>animation_type</dfn> specifies the type of animation applied to the parameter values. When an unknown value of [=animation_type=] is used, parsers compliant to this version of the specification SHALL ignore the [=Parameter Block OBU=] including the [=mix_gain_parameter_data()=].
 
 <dfn noexport>param_data</dfn> uses the AnimatedParameterData function template. Each of the values defined within this instance (start_point_value, end_point_value and control_point_value) is expressed in dB and SHALL be applied to all channels in the rendered [=Audio Element=]. They are stored as 16-bit, signed, two's complement fixed-point values with 8 fractional bits (i.e. Q7.8 in [[!Q-Format]]).
 
--- a/index.bs
+++ b/index.bs
@@ -906,7 +906,7 @@ NOTE: The Ltr and Rtr of 5.1.4ch down-mixed from 7.1.4ch is within the range of 
 - 0: No recon_gain information fields for the [=ChannelGroup=] present in [=recon_gain_info_parameter_data()=].
 - 1: recon_gain information fields for the [=ChannelGroup=] present in [=recon_gain_info_parameter_data()=]. In this case, [=recon_gain_flags=] and [=recon_gain=] fields present.
 
-<dfn noexport>substream_count</dfn> specifies the number of [=Audio Substream=]s. The sum of all [=substream_count]s in this OBU SHALL be the same as [=num_substreams=] in this OBU. It SHALL NOT be set to 0.
+<dfn noexport>substream_count</dfn> specifies the number of [=Audio Substream=]s. The sum of all [=substream_count=]s in this OBU SHALL be the same as [=num_substreams=] in this OBU. It SHALL NOT be set to 0.
 
 <dfn noexport>coupled_substream_count</dfn> specifies the number of referenced [=Audio Substream=]s, each of which are coded as coupled stereo channels.
 
--- a/index.bs
+++ b/index.bs
@@ -576,6 +576,8 @@ class codec_config() {
 - 'fLaC': All coded [=Audio Substream=]s of all [=Audio Element=]s referring to this codec configuration SHALL comply to the specification [[!FLAC]] and the [=decoder_config()=] structure SHALL comply with the constraints given in [[#flac-specific]].
 - 'ipcm': All coded [=Audio Substream=]s of all [=Audio Element=]s referring to this codec configuration SHALL be audio samples for linear PCM (LPCM) and the [=decoder_config()=] structure SHALL comply with the constraints given in [[#lpcm-specific]].
 
+Parsers compliant to this version of the specification SHALL ignore [=Codec Config OBU=]s with an unknown [=codec_id=].
+
 NOTE: 'ipcm' should not be confused with 'lpcm' which is another 4CC to identify codecs in other container formats (e.g. QuickTime).
 
 <dfn noexport>num_samples_per_frame</dfn> indicates the frame length, in samples, of the audio_frame() provided in by audio_frame_obu(). It SHALL NOT be set to zero. If the [=decoder_config()=] structure for a given codec specifies a value for the frame length, the two values SHALL be equal.
@@ -616,13 +618,16 @@ class audio_element_obu() {
     if (param_definition_type == PARAMETER_DEFINITION_RECON_GAIN) {
         ReconGainParamDefinition recon_gain_info;
     }
+    if (param_definition_type > 2) {
+        leb128() param_definition_size;
+    }
   }
 
   if (audio_element_type == CHANNEL_BASED) {
     scalable_channel_layout_config();
   } else if (audio_element_type == SCENE_BASED) {
     ambisonics_config();
-  }  
+  } 
 }
 ```
 
@@ -644,7 +649,7 @@ class ReconGainParamDefinition() extends ParamDefinition() {
 
 <dfn noexport>audio_element_id</dfn> defines an identifier for an [=Audio Element=]. Within an [=IA Sequence=], there SHALL be exactly one non-redundant [=Audio Element OBU=] with a given identifier. [=Mix Presentation=]s that use an [=Audio Element=] refer to this identifier.
 
-<dfn noexport>audio_element_type</dfn> specifies the audio representation of this [=Audio Element=] which is constructed from one or more [=Audio Substream=]s.
+<dfn noexport>audio_element_type</dfn> specifies the audio representation of this [=Audio Element=] which is constructed from one or more [=Audio Substream=]s. Parsers compliant to this version of the specification SHALL ignore [=Audio Element OBU=]s with a reserved [=audio_element_type=].
 
 <pre class = "def">
 audio_element_type: The type of audio representation.
@@ -653,9 +658,9 @@ audio_element_type: The type of audio representation.
   2~7   : Reserved
 </pre>
 
-<dfn value noexport for="audio_element_obu()">codec_config_id</dfn> indicates the identifier for the codec configuration which this [=Audio Element=] refers to.
+<dfn value noexport for="audio_element_obu()">codec_config_id</dfn> indicates the identifier for the codec configuration which this [=Audio Element=] refers to. Parsers compliant to this version of the specification SHALL ignore [=Audio Element OBU=]s with a [=codec_config_id=] identifying an unknown [=codec_id=].
 
-<dfn noexport>num_substreams</dfn> specifies the number of [=Audio Substream=]s that are used to reconstruct this [=Audio Element=].
+<dfn noexport>num_substreams</dfn> specifies the number of [=Audio Substream=]s that are used to reconstruct this [=Audio Element=]. It SHALL NOT be set to 0.
 
 <dfn value noexport for="audio_element_obu()">audio_substream_id</dfn> indicates the identifier for an [=Audio Substream=] which this [=Audio Element=] refers to.
 
@@ -683,6 +688,8 @@ A [=ChannelGroup=] is defined in [[#iamfgeneration]]. The order of the [=Audio S
 - When the highest [=loudspeaker_layout=] of (non-)scalable channel audio is less than or equal to 3.1.2ch, the type PARAMETER_DEFINITION_DEMIXING SHALL NOT be present.
 - When the highest [=loudspeaker_layout=] of scalable channel audio ([=num_layers=] > 1) is greater than 3.1.2ch, the types of both PARAMETER_DEFINITION_DEMIXING and both PARAMETER_DEFINITION_RECON_GAIN SHALL be present.
 - When [=num_layers=] = 1 and [=loudspeaker_layout=] is greater than 3.1.2ch, the type PARAMETER_DEFINITION_DEMIXING MAY be present.
+- Parsers compliant to this version of the specification SHALL ignore a future parameter definition with [=param_definition_type=] = P (where, P > 2).
+- An OBU parser which is conformant with this version of the specification SHALL be able to parse [=param_definition_type=] = P (where, P > 2) and [=param_definition_size=]. And, the OBU Parser SHALL be able to ignore or skip the bytes indicated by [=param_definition_size=].
 
 <table class = "def">
 <tr>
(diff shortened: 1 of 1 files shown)


felicialim (approved):

cconcolato:
I'm ok with the change but just want us to be conscious that this means we will not introduce be able to introduce any backwards incompatible change in existing OBUs, and that we will have to use new OBUs if existing OBUs need to be changed.

sunghee-hwang:
> I'm ok with the change but just want us to be conscious that this means we will not introduce be able to introduce any backwards incompatible change in existing OBUs, and that we will have to use new OBUs if existing OBUs need to be changed.

I agree with you. But it still allows that existing OBU payloads can be updated with backwards incompatible. (e.g. when a reserved field in an OBU payload is used for future version. The current parsers discards or ignores the OBU).

IMHO is as follows:
- Parsers which are compliant with this version of the specification SHALL be able to parser any future OBU header. For this, one way is to fix OBU header structure forever and another way is to introduce extension header by using header_extension_flag (the latter is the way what we have taken)
- Parsers which are compliant with this version of the specification SHALL be able to parser any future OBU payload with a known obu_type as long as the future OBU payload is backward compatible with the current OBU with the same obu_type.
- Parsers which are compliant with this version of the specification SHOULD be able to ignore/discard any future OBU payload with a known obu_type if the future OBU payload is backward incompatible with the current OBU with the same obu_type.

index.bs
```diff
@@ -1328,6 +1341,8 @@ class loudness_info() {
  2~7 : Reserved
 </pre>
 
+When a bitmask for reserved of [=info_type=] is set, parsers compliant to this version of the specification SHOULD ignore the syntaxes defined by the bitmask.
```
felicialim, line 1344: Can we do this without knowing the size of the new syntax? Or should we provide the extension size, as we do elsewhere?

index.bs
```diff
@@ -1574,7 +1591,7 @@ NOTE: The [=Temporal Delimiter OBU=] has an empty payload.
 
 ## Codec Specific ## {#codec-specific}
 
-This section defines codec specific information for Codec_Specific_Info and its coded [=Audio Substream=].
+This section defines codec specific information for Codec_Specific_Info and its coded [=Audio Substream=]. To generate one single coded [=Audio Substream=], mono or stereo coding SHALL be only allowed for this version of the specification.
```
felicialim, line 1594:
```suggestion
This section defines codec specific information for Codec_Specific_Info and its coded [=Audio Substream=]. To generate one single coded [=Audio Substream=], only mono or stereo coding SHALL be allowed for this version of the specification.
```

index.bs
```diff
@@ -683,6 +691,7 @@ A [=ChannelGroup=] is defined in [[#iamfgeneration]]. The order of the [=Audio S
 - When the highest [=loudspeaker_layout=] of (non-)scalable channel audio is less than or equal to 3.1.2ch, the type PARAMETER_DEFINITION_DEMIXING SHALL NOT be present.
 - When the highest [=loudspeaker_layout=] of scalable channel audio ([=num_layers=] > 1) is greater than 3.1.2ch, the types of both PARAMETER_DEFINITION_DEMIXING and both PARAMETER_DEFINITION_RECON_GAIN SHALL be present.
 - When [=num_layers=] = 1 and [=loudspeaker_layout=] is greater than 3.1.2ch, the type PARAMETER_DEFINITION_DEMIXING MAY be present.
+- An OBU parser which is conformant with this version of the specification SHALL be able to parse [=param_definition_type=] = P (where, P > 2) and [=param_definition_size=]. And, the OBU Parser SHALL be able to ignore or skip the bytes indicated by [=param_definition_size=].
```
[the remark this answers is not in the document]
sunghee-hwang: We could. But it depends on what our wish is on whether a future version is backward compatible or not.
For example, when we add a parameter substream for ambisonics in future version. This PR makes it be backward compatible.

index.bs
sunghee-hwang: Yes you are right. It should ignore the following all bytes in the OBU.
Generally, I am thinking how to handle class() which it does not have its size field.

index.bs
```diff
@@ -843,9 +856,9 @@ The IA decoder SHALL select one of one or more channel audios provided by scalab
 
 <dfn noexport>channel_audio_layer_config()</dfn> is a class that provides the information regarding the configuration of [=ChannelGroup=] for scalable channel audio. channel_audio_layer_config(i) provides information regarding the configuration of [=ChannelGroup=] #i.
 
-<dfn noexport>loudspeaker_layout</dfn> indicates the channel layout for the channels to be reconstructed from the precedent [=ChannelGroup=]s and the current [=ChannelGroup=] among [=ChannelGroup=]s for scalable channel audio.
+<dfn noexport>loudspeaker_layout</dfn> indicates the channel layout for the channels to be reconstructed from the precedent [=ChannelGroup=]s and the current [=ChannelGroup=] among [=ChannelGroup=]s for scalable channel audio. When a reserved value for [=loudspeaker_layout=] is used, parsers compliant to this version of the specification SHOULD skip that layer and all subsequent ones.
```
sunghee-hwang, line 859:
```suggestion
<dfn noexport>loudspeaker_layout</dfn> indicates the channel layout for the channels to be reconstructed from the precedent [=ChannelGroup=]s and the current [=ChannelGroup=] among [=ChannelGroup=]s for scalable channel audio. When a reserved value for [=loudspeaker_layout=] is used, parsers compliant to this version of the specification SHOULD skip that layer and all subsequent ones if any.
```

index.bs
sunghee-hwang, line 1344:
```suggestion
When a bitmask for reserved of [=info_type=] is set, parsers compliant to this version of the specification SHOULD ignore all bytes from the first byte of the syntaxes defined by the bitmask to the last byte of the OBU.
```

index.bs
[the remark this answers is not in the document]
felicialim: Ok, I think this is fine given we should define new audio elements instead, if they need the new parameters to be able to be reconstructed correctly.

felicialim (approved):

index.bs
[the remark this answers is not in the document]
sunghee-hwang: Yes right. A future will decide to define a new audio element for that or to make an extended OBU be backward compatible with the current one.

index.bs
```diff
@@ -474,7 +474,7 @@ NOTE: This means that if one of the value is set to the the number of samples (i
 
 <dfn noexport>obu_extension_flag</dfn> indicates whether the [=extension_header_size=] field presents or not. If it set to 0, the [=extension_header_size=] field SHALL NOT be present. Otherwise, the [=extension_header_size=] field SHALL be present.
 
-This flag SHALL be set to 0 for this version of the specification. An OBU parser which is conformant with this version of the specification SHALL be able to parse this flag and [=extension_header_size=].
+This flag SHALL be set to 0 for this version of the specification. An OBU parser which is conformant with this version of the specification SHALL be able to parse this flag and [=extension_header_size=] and SHOULD ignore the extension header.
```
cconcolato, line 477: I don't understand what "SHALL be able to parse this flag" means. Parsing a flag is just reading 1 bit. Also this bit/flag is in the first version of the spec so all parser will parse it. I think the important part is what happens with the extension header. Why is it a should?
```suggestion
This flag SHALL be set to 0 for this version of the specification. An OBU parser which is conformant with this version of the specification SHALL ignore the `extension_header_bytes`.
```
Also the syntax of `obu_header` is missing `extension_header_bytes`

index.bs
```diff
@@ -672,8 +677,11 @@ A [=ChannelGroup=] is defined in [[#iamfgeneration]]. The order of the [=Audio S
 
 
 <dfn noexport>num_parameters</dfn> specifies the number of [=Parameter Substream=]s that are used by the algorithms specified in this [=Audio Element=].
-- When [=audio_element_type=] = 0, this field SHALL be set to 0, 1 or 2.
-- When [=audio_element_type=] = 1, this field SHALL be set to 0.
+- When [=audio_element_type=] = 0, this field SHALL be set to 0, 1 or 2 for this version of the specification.
+- When [=audio_element_type=] = 1, this field SHALL be set to 0 for this version of the specification.
+- Parsers compliant to this version of the specification SHALL be able to parse any value of [=num_parameters=].
+
+NOTE: For a given [=audio_element_type=], a future version of the specification may define a new [=Parameter Substream=] which may be ignored by IA decoders compliant to this version of the specification. In that case, [=num_parameters=] may be set to another value in a future version of [=Audio Element OBU=].
```
cconcolato, line 684: Is it `num_parameters` or is it `param_definition_type` ?

index.bs
```diff
@@ -1289,6 +1300,8 @@ layout_type : Layout type
  - 12: It indicates Mono
  - 13 ~ 15: Reserved
 
+When a reserved value for [=sp_label=] or [=sound_system=] is used, parsers compliant to this version of the specification SHALL ignore this layout() and the following [=loudness_info()=].
```
cconcolato, line 1303: It's unusual to say "when a reserved value [...] is used". It should rather say "when a the value is not supported"

index.bs
```diff
@@ -616,13 +618,16 @@ class audio_element_obu() {
     if (param_definition_type == PARAMETER_DEFINITION_RECON_GAIN) {
         ReconGainParamDefinition recon_gain_info;
     }
+    if (param_definition_type > 2) {
```
sunghee-hwang, line 621:
```suggestion
```

index.bs
```diff
@@ -616,13 +618,16 @@ class audio_element_obu() {
...
+        leb128() param_definition_size;
```
sunghee-hwang, line 622:
```suggestion
```

index.bs
```diff
@@ -616,13 +618,16 @@ class audio_element_obu() {
...
+    }
```
sunghee-hwang, line 623:
```suggestion
```

index.bs
```diff
@@ -699,6 +708,10 @@ A [=ChannelGroup=] is defined in [[#iamfgeneration]]. The order of the [=Audio S
 </tr>
 </table>
 
+<dfn noexport>param_definition_size</dfn> indicates the size in bytes of the parameter definition with [=param_definition_type=] = P (where, P > 2) including this field.
```
sunghee-hwang, line 711:
```suggestion
```

index.bs
```diff
@@ -699,6 +708,10 @@ A [=ChannelGroup=] is defined in [[#iamfgeneration]]. The order of the [=Audio S
...
+
```
sunghee-hwang, line 712:
```suggestion
```

index.bs
```diff
@@ -699,6 +708,10 @@ A [=ChannelGroup=] is defined in [[#iamfgeneration]]. The order of the [=Audio S
...
+NOTE: A future version of the specification may specify a new parameter definition by setting [=param_definition_type=] = P (where, P > 2) and setting the sum of the size of the new parameter definition and the size of this field to [=param_definition_size=].
```
sunghee-hwang, line 713:
```suggestion
```

index.bs
```diff
@@ -699,6 +708,10 @@ A [=ChannelGroup=] is defined in [[#iamfgeneration]]. The order of the [=Audio S
...
+
```
sunghee-hwang, line 714:
```suggestion
```

sunghee-hwang (commented):
It is reverted to be addressed by a new PR

sunghee-hwang:
FYI, I updated the definition of extension_header_size to align with the other definitions for size fields. So, it is not labeled as new_bitstream_required.

index.bs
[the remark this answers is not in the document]
sunghee-hwang: Please consider that it is now addressed by another PR.

index.bs
```diff
@@ -672,8 +676,11 @@ A [=ChannelGroup=] is defined in [[#iamfgeneration]]. The order of the [=Audio S
 
 
 <dfn noexport>num_parameters</dfn> specifies the number of [=Parameter Substream=]s that are used by the algorithms specified in this [=Audio Element=].
-- When [=audio_element_type=] = 0, this field SHALL be set to 0, 1 or 2.
-- When [=audio_element_type=] = 1, this field SHALL be set to 0.
+- When [=audio_element_type=] = 0, this field SHALL be set to 0, 1 or 2 for this version of the specification.
+- When [=audio_element_type=] = 1, this field SHALL be set to 0 for this version of the specification.
+- Parsers compliant to this version of the specification SHALL be able to parse any value of [=num_parameters=].
+
+NOTE: For a given [=audio_element_type=], a future version of the specification may define a new [=Parameter Substream=] which may be ignored by IA decoders compliant to this version of the specification. In that case, [=num_parameters=] may be set to another value in a future version of [=Audio Element OBU=].
```
sunghee-hwang, line 683:
```suggestion
NOTE: For a given [=audio_element_type=], a future version of the specification may define a new [=Parameter Substream=] which may be ignored by IA decoders compliant to this version of the specification. In that case, a new [=param_definition_type=] will defined in a future version of [=Audio Element OBU=].
```

index.bs
```diff
@@ -1289,6 +1295,8 @@ layout_type : Layout type
  - 12: It indicates Mono
  - 13 ~ 15: Reserved
 
+When a reserved value for [=sp_label=] or [=sound_system=] is used, parsers compliant to this version of the specification SHALL ignore this layout() and the following [=loudness_info()=].
```
sunghee-hwang, line 1298:
```suggestion
When a value for [=sp_label=] or [=sound_system=] is not supported, parsers compliant to this version of the specification SHALL ignore this layout() and the following [=loudness_info()=].
```

sunghee-hwang (commented):
To address the comments

index.bs
```diff
@@ -893,14 +901,13 @@ NOTE: The Ltr and Rtr of 5.1.4ch down-mixed from 7.1.4ch is within the range of
...
 
 <dfn noexport>coupled_substream_count</dfn> specifies the number of referenced [=Audio Substream=]s, each of which are coded as coupled stereo channels.
 
-Mono or stereo coding SHALL be only allowed for the version of this specification.
-- Each pair of coupled stereo channels in the same [=ChannelGroup=] SHALL be coded as stereo mode to generate one single coded [=Audio Substream=] and each of non-coupled channels in the same [=ChannelGroup=] SHALL be coded as mono mode to generate one single coded [=Audio Substream=].
-	- <dfn noexport>Coupled stereo channels</dfn>: L/R, Ls/Rs, Lss/Rss, Lrs/Rrs, Ltf/Rtf, Ltb/Rtb
-	- <dnf noexport>Non-coupled channels</dfn>: C, LFE, L
+Each pair of coupled stereo channels in the same [=ChannelGroup=] SHALL be coded in stereo mode to generate one single coded [=Audio Substream=] and each of the non-coupled channels in the same [=ChannelGroup=] SHALL be coded in mono mode to generate one single coded [=Audio Substream=].
+- <dfn noexport>Coupled stereo channels</dfn>: L/R, Ls/Rs, Lss/Rss, Lrs/Rrs, Ltf/Rtf, Ltb/Rtb
+- <dnf noexport>Non-coupled channels</dfn>: C, LFE, L
```
felicialim, line 910: Drop L?
If so, the sentence below also needs to be changed to delete "followed by the other one":
"Center channel comes first and followed by LFE and followed by the other one."

index.bs
```diff
@@ -1424,7 +1436,7 @@ class mix_gain_parameter_data() {
 
 <b>Semantics</b>
 
-<dfn noexport>animation_type</dfn> specifies the type of animation applied to the parameter values.
+<dfn noexport>animation_type</dfn> specifies the type of animation applied to the parameter values. When an unknown value of [=animation_type=] is used, parsers compliant to this version of the specification SHALL ignore the [=Parameter Block OBU=] including the [=mix_gain_parameter_data()=].
```
felicialim, line 1439:
```suggestion
<dfn noexport>animation_type</dfn> specifies the type of animation applied to the parameter values. When an unknown value of [=animation_type=] is used, parsers compliant to this version of the specification SHALL ignore the [=Parameter Block OBU=] that contains this [=mix_gain_parameter_data()=].
```

index.bs
felicialim: Ignore: L is the left channel that is needed when we have a 2-layer scalable channel for stereo, with the first layer as Mono

cconcolato (approved):

PR #504, opened 2023-06-22
Fix Bezier: when alpha==0

felicialim:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/504.html" title="Last updated on Jun 26, 2023, 5:57 PM UTC (12d7664)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/504/3318721...12d7664.html" title="Last updated on Jun 26, 2023, 5:57 PM UTC (12d7664)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2272,6 +2272,7 @@ beta = 2 * (n1 - n0),
 gamma = n0 - n.
 ```
 
+If the value of <code>alpha</code> is 0, it MUST be set to a near-zero value instead, for example 0.001.
 
 ## Post Processing ## {#processing-post}
 
--- a/index.bs
+++ b/index.bs
@@ -2272,7 +2272,7 @@ beta = 2 * (n1 - n0),
 gamma = n0 - n.
 ```
 
-If the value of <code>alpha</code> is 0, it SHALL be set to a near-zero value instead, for example 0.001.
+If the value of <code>alpha</code> is 0, it SHALL be set to a near-zero value instead.
 
 ## Post Processing ## {#processing-post}
 
--- a/index.bs
+++ b/index.bs
@@ -2272,7 +2272,7 @@ beta = 2 * (n1 - n0),
 gamma = n0 - n.
 ```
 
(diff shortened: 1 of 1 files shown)


felicialim:
When `control_point_relative_time == 0.5`, `n1 = 0.5 * n2` and `alpha = 0`, which leads to a divide-by-zero when calculating `a` (`a = 0 / 0`).

index.bs
```diff
@@ -2272,6 +2272,7 @@ beta = 2 * (n1 - n0),
 gamma = n0 - n.
 ```
 
+If the value of <code>alpha</code> is 0, it MUST be set to a near-zero value instead, for example 0.001.
```
sunghee-hwang, line 2275:
```suggestion
If the value of <code>alpha</code> is 0, it SHALL be set to a near-zero value instead, for example 0.001.
```

sunghee-hwang (approved):

jwcullen:
Maybe instead it is better to define `a` as a piecewise function?

When alpha != 0:
`a = (-beta + sqrt(beta^2 - 4 * alpha * gamma)) ÷ (2 * alpha)`
When alpha = 0:
`a = -gamma ÷ beta`

I think latter function is what the math works out to be when calculating the limit as `alpha` approaches 0 of the former function. Defining it as a piecewise function like this means everyone will get the same result instead of different implementations choosing different small values of `alpha`.

felicialim:
> Maybe instead it is better to define `a` as a piecewise function?
> 
> When alpha != 0: `a = (-beta + sqrt(beta^2 - 4 * alpha * gamma)) ÷ (2 * alpha)` When alpha = 0: `a = -gamma ÷ beta`
> 
> I think latter function is what the math works out to be when calculating the limit as `alpha` approaches 0 of the former function. Defining it as a piecewise function like this means everyone will get the same result instead of different implementations choosing different small values of `alpha`.

Done.

jwcullen (approved):

tdaede (approved):

PR #505, opened 2023-06-23
Fix #499, profile_name -> primary_profile

### LINKED ISSUE #499: Different name for profile_name
We have profile_name and profile_compatible, there might be a better name for profile_name.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/505.html" title="Last updated on Jun 28, 2023, 12:18 AM UTC (6c160ec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/505/3318721...6c160ec.html" title="Last updated on Jun 28, 2023, 12:18 AM UTC (6c160ec)">Diff</a>


index.bs
```diff
@@ -534,17 +534,16 @@ class ia_sequence_header_obu() {
...
 
-<dfn noexport>profile_name</dfn> indicates that this [=IA Sequence=] is able to be processed by IA decoders that support the profile indicated in this field. The IA decoders SHALL be able to parse all OBUs explicitly listed in that profile and can still encounter reserved OBUs that it SHOULD skip and it is acceptable to do so. This allows future versions of the specification to define new profiles that can be backwards compatible to old profiles.
-- 0: Simple Profile
-- 1: Base Profile
-- 2~255: Reserved
+<dfn noexport>primary_profile</dfn> indicates that this [=IA Sequence=] is able to be processed by IA decoders that support the profile indicated in this field. The IA decoders SHALL be able to parse all OBUs explicitly listed in that profile and can still encounter reserved OBUs that it SHOULD skip and it is acceptable to do so. This allows future versions of the specification to define new profiles that can be backwards compatible to old profiles. Parsers compliant to this version of the specification SHALL discard this OBU with [=primary_profile=] = a value of reserved and all subsequent OBUs until encounting the next [=IA Sequence Header OBU=].
```
cconcolato, line 537:
```suggestion
<dfn noexport>primary_profile</dfn> indicates the primary profile this [=IA Sequence=] complies to. Parsers compliant to this version of the specification SHALL discard the [=IA Sequence=] when this value is not one they support.
```
The rest of the text should go to the section about profiles:
``IA decoders SHALL be able to parse all OBUs explicitly listed in a profile they support. They can still encounter reserved OBUs that they SHOULD skip. This allows future versions of the specification to define new profiles that can be backwards compatible to old profiles.
```

File: index.bs [hunk 2]
```diff
@@ -534,17 +534,16 @@ class ia_sequence_header_obu() {
...
+<dfn noexport>additional_profile</dfn> indicates an additional profile of this specification to which this [=IA Sequence=] complies. If an [=IA Sequence=] only complies with the [=primary_profile=] profile, this field SHALL be set to same [=primary_profile=] value. It SHALL be set to greater than or equal to the value of [=primary_profile=]. Parsers compliant to this version of the specification SHALL discard this OBU with the value of [=additional_profile=] < the value of [=primary_value=] and all subsequent OBUs until encounting the next [=IA Sequence Header OBU=].
```
cconcolato, line 544: Why add the restriction that it shall be greater? It depends on how profiles are defined. I suggest using SHOULD and remove the last sentence.

index.bs
```diff
@@ -534,17 +534,16 @@ class ia_sequence_header_obu() {
...
+
+NOTE: For example, an [=IA Sequence=] with [=primary_profile=] = 0 and [=additional_profile=] = 1 means that the [=IA Sequence=] complies to Base profile of specification and IA decoder to support Simple profile can playback the [=IA Sequence=]. When a future profile is defined with [=additional_profile=] = 2 as called Enhanced profile, an [=IA Sequence=] with [=primary_profile=] = 1 and [=additional_profile=] = 2 will mean that the [=IA Sequence=] will comply with Enhanced profile and IA decoder to support Base profile of specification will be able to playback it by ignoring unknown OBUs and/or unknown syntaxes if present.
```
cconcolato, line 546: The example of Simple/Base is not the best choice because both profiles are defined in the same version of the spec. It is clear that a Simple bitstream can be decoded by a Base decoder.
The second example should not use a real name like 'Enhanced'. 
```suggestion
NOTE: If a future version defines a new profile, e.g. HypotheticalProfile, that is backwards compatible with the Base profile, for example by defining new OBUs that would be ignored by the Base-compatible parser, a IA writer can decide to set the [=primary_profile=] to 'Base' while setting the [=additional_profile=] to 'HypotheticalProfile'. This way an old processor will know it can parse and produce an acceptable rendering, while a new processor still know it can produce a better result because it will not ignore the additional features.
```

index.bs
```diff
@@ -1684,12 +1683,12 @@ Restrictions on the [=IA Sequence=]:
 
 - There SHALL be only one unique [=Audio Element OBU=].
 - There SHALL NOT be any [=Temporal Delimiter OBU=]s present.
-- [=profile_compatible=] SHALL be set to 0 to indicate that the [=IA Sequence=] complies to this profile.
-- [=profile_name=] SHALL be set to 0.
+- [=additional_profile=] SHALL be set to 0 to indicate that the [=IA Sequence=] complies to this profile.
```
cconcolato, line 1686: The profile definition should not have any restrictions on `additional_profile` because otherwise future streams cannot be backwards compatible with this profile.

index.bs
```diff
@@ -1708,12 +1707,12 @@ Restrictions on [=IA Sequence=]:
 		- Scene-based [=Audio Element=] + Channel-based [=Audio Element=] having [=num_layers=] = 1.
 		- Scene-based [=Audio Element=] + Channel-based [=Audio Element=] having [=num_layers=] > 1.
 - There MAY be [=Temporal Delimiter OBU=]s present. If present, the first OBU of every [=Temporal Unit=] SHALL be [=Temporal Delimiter OBU=].
-- [=profile_compatible=] SHALL be set to 1 to indicate that the [=IA Sequence=] complies to this profile.
-- [=profile_name=] MAY be set to 0 or 1.
+- [=additional_profile=] SHALL be set to 1 to indicate that the [=IA Sequence=] complies to this profile.
```
cconcolato, line 1710: Same comment as above.

cconcolato (changes requested):

index.bs
tdaede: The concern is that if profile compatibility becomes an arbitrary tree, we might need more than 2 profile fields. But we can be careful to not create that sort of situation.

We can remove the restriction on "greater than" to solve this (as Cyril commented).

index.bs
tdaede: We'll remove this restriction.

index.bs
sunghee-hwang: > The example of Simple/Base is not the best choice because both profiles are defined in the same version of the spec. It is clear that a Simple bitstream can be decoded by a Base decoder. The second example should not use a real name like 'Enhanced'.

```suggestion
NOTE: If a future version defines a new profile, e.g. HypotheticalProfile, that is backwards compatible with the Base profile, for example by defining new OBUs that would be ignored by the Base-compatible parser, a IA writer can decide to set the [=primary_profile=] to 'Base' while setting the [=additional_profile=] to 'HypotheticalProfile'. This way an old processor will know it can parse and produce an acceptable rendering, while a new processor still knows it can produce a better result because it will not ignore the additional featurest.
```

index.bs
```diff
@@ -534,17 +534,16 @@ class ia_sequence_header_obu() {
...
+<dfn noexport>additional_profile</dfn> indicates an additional profile of this specification to which this [=IA Sequence=] complies. If an [=IA Sequence=] only complies with the [=primary_profile=] profile, this field SHALL be set to same [=primary_profile=] value. It SHALL be set to greater than or equal to the value of [=primary_profile=]. Parsers compliant to this version of the specification SHALL discard this OBU with the value of [=additional_profile=] < the value of [=primary_value=] and all subsequent OBUs until encounting the next [=IA Sequence Header OBU=].
```
sunghee-hwang, line 544:
```suggestion
<dfn noexport>additional_profile</dfn> indicates an additional profile of this specification to which this [=IA Sequence=] complies. If an [=IA Sequence=] only complies with the [=primary_profile=] profile, this field SHALL be set to same [=primary_profile=] value. Parsers compliant to this version of the specification SHALL discard this OBU with the value of [=additional_profile=] < the value of [=primary_value=] and all subsequent OBUs until encounting the next [=IA Sequence Header OBU=].
```

index.bs
```diff
@@ -534,17 +534,16 @@ class ia_sequence_header_obu() {
...
 - 0: Simple Profile
 - 1: Base Profile
 - 2~255: Reserved
 
-NOTE: For example, an [=IA Sequence=] with [=profile_name=] = 0 and [=profile_compatible=] = 1 means that the [=IA Sequence=] complies to Base profile of specification and IA decoder to support Simple profile can playback the [=IA Sequence=]. When a future profile is defined with [=profile_compatible=] = 2 as called Enhanced profile, an [=IA Sequence=] with [=profile_name=] = 1 and [=profile_compatible=] = 2 will mean that the [=IA Sequence=] will comply with Enhanced profile and IA decoder to support Base profile of specification will be able to playback it by ignoring unknown OBUs and/or unknown syntaxes if present.
+<dfn noexport>additional_profile</dfn> indicates an additional profile of this specification to which this [=IA Sequence=] complies. If an [=IA Sequence=] only complies with the [=primary_profile=] profile, this field SHALL be set to same [=primary_profile=] value. Parsers compliant to this version of the specification SHALL discard this OBU with the value of [=additional_profile=] < the value of [=primary_value=] and all subsequent OBUs until encounting the next [=IA Sequence Header OBU=].
```
sunghee-hwang, line 544:
```suggestion
<dfn noexport>additional_profile</dfn> indicates an additional profile of this specification to which this [=IA Sequence=] complies. If an [=IA Sequence=] only complies with the [=primary_profile=] profile, this field SHALL be set to same [=primary_profile=] value.
```

cconcolato (approved):

PR #506, opened 2023-06-23
Fix #460, updated codecs parameter string

### LINKED ISSUE #460: Consider adding profile_compatible to codecs string.
https://github.com/AOMediaCodec/iamf/pull/453#discussion_r1218673740

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/506.html" title="Last updated on Jun 28, 2023, 10:29 PM UTC (8686437)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/506/3318721...8686437.html" title="Last updated on Jun 28, 2023, 10:29 PM UTC (8686437)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1875,20 +1875,21 @@ DASH and other applications require defined values for the 'Codecs' parameter sp
 	iamf.IAMF-specific-needs.ipcm
 ```
 
-<b>IAMF-specific-needs</b> SHALL be <b>PC</b> as follows:
-- <dfn noexport>PC</dfn> is three digits within the range 0 to 255 and represents that IA decoder supports the profile of the specification.
+<b>IAMF-specific-needs</b> SHALL be <b>PP</b>.<b>AP</b> as follows:
+- <dfn noexport>PP</dfn> is three digits within the range 0 to 255 and represents that an IA decoder supports the profile of the specification or the [=primary_profile=] of an [=IA Sequence=].
+- <dfn noexport>AP</dfn> is three digits within the range 0 to 255 and represents that an IA decoder supports the profile of the specification or the [=additional_profile=] of an [=IA Sequence=].
 
 For example, for this version of the specification
-- The codecs parameter string of OPUS for the simple profile:
+- The codecs parameter string of OPUS for [=primary_profile=] = simple profile and [=additional_profile=] = simple profile:
 
 ```
-	iamf.000.Opus
+	iamf.000.000.Opus
 ```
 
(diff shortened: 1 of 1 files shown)


tdaede (approved):

cconcolato:
`Codecs` should be spelled lower case `codecs`

cconcolato:
The sentence 
```
The codecs parameter string for the AOM IA codec SHALL be
```
uses "AOM IA codec" which is not really defined.

cconcolato:
Also, generally, I think the text should be improved. RFC 6381 (and other derived specs) indicates that the parameter is made of elements ("Each value consists of one or more dot-separated elements"). The ISOBMFF part says:
> For the ISO Base Media File Format, and the QuickTime movie file
   format, the first element of a 'codecs' parameter value is a sample
   description entry four-character code as registered by the MP4
   Registration Authority [[MP4RA](https://datatracker.ietf.org/doc/html/rfc6381#ref-MP4RA)].  Values are case sensitive.

This specification should say:
- Per RFC6381 and ISOBMFF, the first element of the codecs parameter string is `iamf`
- The second element indicates the primary_profile
- The 3rd element indicates the additional_profile
- The 4th element and any additional elements, if any, shall be the elements of the codecs parameter string if that stream was carried in its own track (i.e. not encapsulated in IAMF).

cconcolato (changes requested):
See the "conversation" tab for comments

cconcolato:
> * The 4th element and any additional elements, if any, shall be the elements of the codecs parameter string if that stream was carried in its own track (i.e. not encapsulated in IAMF).

In particular, for AAC the spec says:
> iamf.IAMF-specific-needs.mp4a.40.2

but `40.2` is a specific example. There could be other values like `40.5`.

sunghee-hwang:
> > * The 4th element and any additional elements, if any, shall be the elements of the codecs parameter string if that stream was carried in its own track (i.e. not encapsulated in IAMF).
> 
> In particular, for AAC the spec says:
> 
> > iamf.IAMF-specific-needs.mp4a.40.2
> 
> but `40.2` is a specific example. There could be other values like `40.5`.

I know that '40.2' stands for AAC-LC. What does '40.5' stand for?

tdaede:
With the new way the example is written, it no longer specifies 40.2 explicitly, though I think in practice, only 40.2 would be used as the spec is limited to AAC-LC.

PR #508, opened 2023-06-23
Fix #491, deleting rendering_config()

### LINKED ISSUE #491: Rendering config extensibility
Because the rendering config is empty, its size is not known and decoders cannot skip over it. Because it is not backwards compatible, we could remove it entirely from this version of the spec. A version of the specification could use a new OBU.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/508.html" title="Last updated on Jun 23, 2023, 11:45 PM UTC (fd1ad44)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/508/3318721...fd1ad44.html" title="Last updated on Jun 23, 2023, 11:45 PM UTC (fd1ad44)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1015,8 +1015,7 @@ class mix_presentation_obu() {
       leb128() audio_element_id;
       for (i = 0; i < count_label; i++) {
         mix_presentation_element_annotations();
-      }
-      rendering_config();
+      }    
       element_mix_config();
     }
     output_mix_config();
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2103,6 +2087,8 @@ When an [=IA Sequence=] contain multiple [=Mix Presentation=]s, the IA parser SH
 
 ### Rendering an Audio Element ### {#processing-mixpresentation-rendering}
 
+During playback, the rendering of each [=Audio Element=] SHOULD be applied before the processing according to [=element_mix_config()=].
```
cconcolato, line 2090: This sentence is already covered by the 2nd paragraph at https://aomediacodec.github.io/iamf/#processing-mixpresentation

index.bs
```diff
@@ -2103,6 +2087,7 @@ When an [=IA Sequence=] contain multiple [=Mix Presentation=]s, the IA parser SH
 
 ### Rendering an Audio Element ### {#processing-mixpresentation-rendering}
 
+
```
sunghee-hwang, line 2090:
```suggestion
```

PR #533, opened 2023-06-27
Fix #518, Clarify use of AudioSampleEntry channelcount and samplereate

### LINKED ISSUE #518: Clarify use of AudioSampleEntry channelcount and samplereate
A sentence in the spec says:
> The [channelcount](https://www.iso.org/standard/68960.html#) and [samplerate](https://www.iso.org/standard/68960.html#) fields of [AudioSampleEntry](https://www.iso.org/standard/68960.html#) are unused.

What does unused mean? 
The spec should recommend a value and should mandate that reader ignore the fields.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/533.html" title="Last updated on Jun 27, 2023, 7:38 AM UTC (d35b444)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/533/3318721...d35b444.html" title="Last updated on Jun 27, 2023, 7:38 AM UTC (d35b444)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1827,7 +1827,7 @@ class IASampleEntry extends AudioSampleEntry('iamf') {
 }
 ```
 
-The [=channelcount=] and [=samplerate=] fields of [=AudioSampleEntry=] are unused.
+The [=channelcount=] and [=samplerate=] fields of [=AudioSampleEntry=] are set to 1 and to indicate 48000, respectively. Parsers SHALL ignore these two fields.
 
 None of [=AudioSampleEntry=]'s optional boxes SHALL be present.
 
--- a/index.bs
+++ b/index.bs
@@ -1827,7 +1827,9 @@ class IASampleEntry extends AudioSampleEntry('iamf') {
 }
 ```
 
-The [=channelcount=] and [=samplerate=] fields of [=AudioSampleEntry=] are set to 1 and to indicate 48000, respectively. Parsers SHALL ignore these two fields.
+The [=channelcount=] field of [=AudioSampleEntry=] SHALL be set to 0. 
+The [=samplerate=] field of [=AudioSampleEntry=] SHALL be set to the max sample rate of all the [=Audio Substreams=] in the [=IA Sequence=].
+Parsers SHALL ignore these two fields.
 
 None of [=AudioSampleEntry=]'s optional boxes SHALL be present.
 
--- a/index.bs
+++ b/index.bs
@@ -1828,7 +1828,7 @@ class IASampleEntry extends AudioSampleEntry('iamf') {
 ```
 
 The [=channelcount=] field of [=AudioSampleEntry=] SHALL be set to 0. 
-The [=samplerate=] field of [=AudioSampleEntry=] SHALL be set to the max sample rate of all the [=Audio Substreams=] in the [=IA Sequence=].
+The [=samplerate=] field of [=AudioSampleEntry=] SHALL be set to 0. There SHALL be no [=SamplingRateBox=].
 Parsers SHALL ignore these two fields.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1827,7 +1827,7 @@ class IASampleEntry extends AudioSampleEntry('iamf') {
 }
 ```
 
-The [=channelcount=] and [=samplerate=] fields of [=AudioSampleEntry=] are unused.
+The [=channelcount=] and [=samplerate=] fields of [=AudioSampleEntry=] are set to 1 and to indicate 48000, respectively. Parsers SHALL ignore these two fields.
```
cconcolato, line 1830: ISOBMFF has the value 0 for channelcount that means "inapplicable/unknown". It seems preferable.
It might be confusing to set samplerate to a value (48000) if the underlying codec operates at a different rate. 

```suggestion
The [=channelcount=] field of [=AudioSampleEntry=] SHALL be set to 0. 
The [=samplerate=] field of [=AudioSampleEntry=] SHALL be set to the max sample rate of all the [=Audio Substreams=] in the [=IA Sequence=].
Parsers SHALL ignore these two fields.
```

sunghee-hwang: > 
'samplerate' 32bits (16.16) of 'AudioSampleEntry' can indicate maximum 65535.65535Hz.

So, how about this?


```suggestion
The [=channelcount=] field of [=AudioSampleEntry=] SHALL be set to 0. 
The [=samplerate=] field of [=AudioSampleEntry=] SHALL be set to the max sample rate of all the [=Audio Substreams=] in the [=IA Sequence=] if the max sample rate is less than or equal to 48000. Otherwise, it SHALL be set to 0.
Parsers SHALL ignore these two fields.
```

cconcolato: Why does it have to be less than or equal to 48000?

sunghee-hwang: In most cases, I expect that the sample rate would be less than or equal to 48000hz.

cconcolato: No.

I just don't understand why set it to 48k if <= 48k and 0 otherwise.

sunghee-hwang: Then, how to set if the maximum sample rate is for example 96000Hz.?

cconcolato: I see your concern. In that case, yes we should recommend using the SampleRateBox.

cconcolato: image

This is from the ISOBMFF spec. We would fall into the 2) case.

sunghee-hwang: What we need to do is silence?

index.bs
```diff
@@ -1827,7 +1827,9 @@ class IASampleEntry extends AudioSampleEntry('iamf') {
 }
 ```
 
-The [=channelcount=] and [=samplerate=] fields of [=AudioSampleEntry=] are unused.
+The [=channelcount=] field of [=AudioSampleEntry=] SHALL be set to 0. 
+The [=samplerate=] field of [=AudioSampleEntry=] SHALL be set to the max sample rate of all the [=Audio Substreams=] in the [=IA Sequence=].
+Parsers SHALL ignore these two fields.
```
sunghee-hwang, line 1832:
```suggestion
The [=samplerate=] field of [=AudioSampleEntry=] SHALL be set to the max sample rate of all the [=Audio Substreams=] in the [=IA Sequence=]. If the max sample rate greater than the value can be represented in the [=samplerate=] field, then [=SamplingRateBox=] is present in [=AudioSampleEntry=] to document the actual sample rate.
Parsers SHALL ignore these two fields and [=SamplingRateBox=].
```

sunghee-hwang (commented):
To be align with ISBOMFF spec

PR #535, opened 2023-06-27
Fix #530, Consider an extension of loudness info()

### LINKED ISSUE #530: Consider an extension of loudness info()

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/535.html" title="Last updated on Jun 27, 2023, 9:15 PM UTC (e0cca1e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/535/3318721...e0cca1e.html" title="Last updated on Jun 27, 2023, 9:15 PM UTC (e0cca1e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1314,6 +1314,12 @@ class loudness_info() {
       signed int (16) anchored_loudness;
     }
   }
+  for (i = 2; i < 8; i++) {
+    if (info_type & 2^i)) {
+      leb128() info_type_size;
+      unsigned int (8*info_type_size) info_type_bytes;
+    }
+  }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


cconcolato (commented):
An alternative instead of defining reserved bytes for each flag is to reserve a single block and later split it when more flags are defined.

The semantics should indicate that these bytes shall be ignored by parsers compliant to this version of the specification.

index.bs
```diff
@@ -1323,9 +1328,9 @@ class loudness_info() {
```
cconcolato, line 1328: The sentence above says:
> info_type is a bitmask that specifies the type of OPTIONAL loudness information provided. The bits are set as follows, where the first bit is the LSB:

I don't understand why it says 'OPTIONAL'. Can we remove this word?

index.bs
```diff
@@ -1314,6 +1314,11 @@ class loudness_info() {
       signed int (16) anchored_loudness;
     }
   }
+  if (info_type & 4 or info_typ & 8 or info_typ & 16 or info_typ & 32 or info_typ & 64 or info_typ & 128)) {
```
felicialim, line 1317:
```suggestion
  if (info_type & 0b11111100 > 0)) {
```

index.bs
```diff
@@ -1314,6 +1314,11 @@ class loudness_info() {
       signed int (16) anchored_loudness;
     }
   }
+  if (info_type & 0b11111100 > 0) {
```
tdaede, line 1317: editorial nit: this is equivalent to info_type > 3

index.bs
tdaede: We can maybe say "additional loudness information". I think it's to distinguish from integrated_loudness and digital_peak, as those are always provided.

cconcolato (approved):

PR #558, opened 2023-06-29
Fix #511, explicitly limit decoded leb128() to 32 bits

### LINKED ISSUE #511: Specify restriction on leb128 explicitly to be 32 bits max.
We could potentially copy the leb128 definition in from the AV1 spec and drop the normative reference, or at least add the limit explicitly to our definition section.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/558.html" title="Last updated on Jul 4, 2023, 9:29 PM UTC (9c07722)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/558/9e3f2e5...9c07722.html" title="Last updated on Jul 4, 2023, 9:29 PM UTC (9c07722)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2603,12 +2603,6 @@ All of syntax elements conform to [=Syntactic Description Language=] specified i
 
  <b>syntaxName</b> is an unsigned integer which is encoded by <b>leb128()</b>. Its size is limited to 32 bits.
   
- <b>sleb128()</b> <b>syntaxName</b>
-
- <b>sleb128()</b> indicates the type of a signed integer. To encode the following a signed integer <b>syntaxName</b>, it first starts with an N-bit two's complement representation of the signed integer, where N is a multiple of 7, the N-bit representation is broken into groups as for the <b>leb128()</b> encoding.
-
- <b>syntaxName</b> is a signed integer which is encoded by <b>sleb128()</b>. Its size is limited to 32 bits.
- 
  <b>string</b> <b>syntaxName</b>
 
 <b>string</b> indicates a null-terminated (i.e. ending at the first byte set to 0x00), UTF-8 encoded as defined in [[!RFC3629]] and whose length SHALL be limited to 128 bytes. 
--- a/index.bs
+++ b/index.bs
@@ -2599,7 +2599,7 @@ All of syntax elements conform to [=Syntactic Description Language=] specified i
 
  <b>leb128()</b> <b>syntaxName</b>
  
- <b>leb128()</b> indicates the type of an unsigned integer. To encode the following unsigned integer <b>syntaxName</b>, it first represents the integer in binary with an N-bit representation, where N ia a multiple of 7. Then break the integer up into groups of 7 bits. Output one encoded byte for each 7 bits group, from least significant to most significant group. Each byte will have the group in its 7 least significant bits. Set the most significant bit on each byte except the last byte.
+ <b>leb128()</b> indicates the type of an unsigned integer. To encode the following unsigned integer <b>syntaxName</b>, it first represents the integer in binary with an N-bit representation, where N is a multiple of 7. Then break the integer up into groups of 7 bits. Output one encoded byte for each 7 bits group, from least significant to most significant group. Each byte will have the group in its 7 least significant bits. Set the most significant bit on each byte except the last byte.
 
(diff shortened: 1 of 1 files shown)


jwcullen (changes requested):
Two comments

- Perhaps remove `sleb128`. It is not used elsewhere in the spec and it was dropped when Sync OBU was dropped.
- We used to define `leb128` from [AV1](https://aomediacodec.github.io/av1-spec/#leb128). AV1 has a note which allows multiple ways to encode the same value. For example 0 can be encoded as 0x00 or 0x8000., or 0x800000...

_Note: There are multiple ways of encoding the same value depending on how many leading zero bits are encoded. There is no requirement that this syntax descriptor uses the most compressed representation. This can be useful for encoder implementations by allowing a fixed amount of space to be filled in later when the value becomes known._

I think the new definition drops the ability to a `leb128` to be encoded multiple ways as in AV1. It would force 0 to always be encoded as 0x00 for example.

I think the flexibility of allowing multiple representations could have some value in the future. Perhaps some implementations would prefer to allocate fixed sizes for their leb128 values instead of using the minimal representation.

Since we were pulling in the definition from AV1 before this shouldn't be a bitstream change.

index.bs
```diff
@@ -2599,15 +2599,15 @@ All of syntax elements conform to [=Syntactic Description Language=] specified i
 
  <b>leb128()</b> <b>syntaxName</b>
  
- <b>leb128()</b> indicates the type of an unsigned integer. It indicates the following unsigned integer <b>syntaxName</b> is encoded by [=leb128()=] specified in [[!AV1-Convention]].
- 
- <b>syntaxName</b> is an unsigned integer which is encoded by [=leb128()=] specified in [[!AV1-Convention]].
- 
+ <b>leb128()</b> indicates the type of an unsigned integer. To encode the following unsigned integer <b>syntaxName</b>, it first represents the integer in binary. Then zero extends the integer up to a multiple of 7 bits (such that if the number is non-zero, the most significant 7 bits are not all 0). Break the integer up into groups of 7 bits. Output one encoded byte for each 7 bits group, from least significant to most significant group. Each byte will have the group in its 7 least significant bits. Set the most significant bit on each byte except the last byte. The integer zero is encoded as a single byte 0x00.
```
sunghee-hwang, line 2602:
```suggestion
 <b>leb128()</b> indicates the type of an unsigned integer. To encode the following unsigned integer <b>syntaxName</b>, it first represents the integer in binary with an N-bit representation, where N ia a multiple of 7. Then break the integer up into groups of 7 bits. Output one encoded byte for each 7 bits group, from least significant to most significant group. Each byte will have the group in its 7 least significant bits. Set the most significant bit on each byte except the last byte.
```

index.bs
```diff
@@ -2599,15 +2599,15 @@ All of syntax elements conform to [=Syntactic Description Language=] specified i
 
  <b>leb128()</b> <b>syntaxName</b>
  
- <b>leb128()</b> indicates the type of an unsigned integer. It indicates the following unsigned integer <b>syntaxName</b> is encoded by [=leb128()=] specified in [[!AV1-Convention]].
- 
- <b>syntaxName</b> is an unsigned integer which is encoded by [=leb128()=] specified in [[!AV1-Convention]].
- 
+ <b>leb128()</b> indicates the type of an unsigned integer. To encode the following unsigned integer <b>syntaxName</b>, it first represents the integer in binary with an N-bit representation, where N ia a multiple of 7. Then break the integer up into groups of 7 bits. Output one encoded byte for each 7 bits group, from least significant to most significant group. Each byte will have the group in its 7 least significant bits. Set the most significant bit on each byte except the last byte.
```
sunghee-hwang, line 2602:
```suggestion
 <b>leb128()</b> indicates the type of an unsigned integer. To encode the following unsigned integer <b>syntaxName</b>, it first represents the integer in binary with an N-bit representation, where N is a multiple of 7. Then break the integer up into groups of 7 bits. Output one encoded byte for each 7 bits group, from least significant to most significant group. Each byte will have the group in its 7 least significant bits. Set the most significant bit on each byte except the last byte.
```

jwcullen (approved):

tdaede:
We will re-add the note in a follow up PR.

jwcullen:
Approved as limiting to 32 bits and rewording to remove direct reference to AV1.

PR #567, opened 2023-07-04
Fix #540, clarification on IA Sample/Descriptors relationship

### LINKED ISSUE #540: IA Sample format question
Can a sample hold an updated [Audio Element OBU](https://aomediacodec.github.io/iamf/#audio-element-obu) or a [Mix Presentation OBU](https://aomediacodec.github.io/iamf/#mix-presentation-obu) which is different from the one which is stored in the sample entry?

If yes, do we need additional restrictions and/or signaling which would indicate how this information appears in samples?
One way to signal could be the use of dedicated 4CCs similar how it's done in MPEG to signal in-band vs out-of-band config information. 

If this is not allowed then this issue can be closed and the only request would be to clarify the sample format definition to make it more clear how the sample is defined.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/567.html" title="Last updated on Jul 4, 2023, 10:55 PM UTC (5dc01cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/567/9e3f2e5...5dc01cc.html" title="Last updated on Jul 4, 2023, 10:55 PM UTC (5dc01cc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1864,7 +1864,7 @@ The result of encapsulating an [=IA Sequence=] into an [[!ISOBMFF]] file is as f
 
 - If there are audio samples to be trimmed at the start or at the end, then 'edts' and 'elst' boxes SHALL be present to reflect the trimming status.
 - Sample Entry
-	- For an [=IA Sample=], the [=Descriptors=] required for processing the samples SHALL be the [=configOBUs=] of the associated sample entry. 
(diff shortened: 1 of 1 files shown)


tdaede (approved):

tdaede:
I think this solves the issue though the text could be clearer. I can propose something later.

PR #608, opened 2023-07-09
Fix #554, Recommended threshold values for recon_gain

### LINKED ISSUE #554: Are thresholds of -80dB and -6dB for computing recon gain mandated or recommended?
In the section about [recon gain generation](https://aomediacodec.github.io/iamf/#iamfgeneration-scalablechannelaudio-recongaingeneration), two thresholds are mentioned:

> If 10*log10(level Ok / maxL^2) is less than the first threshold value (e.g. **-80dB**), ...

> If 10*log10(level Ok / level Mk ) is less than the second threshold value (e.g. **-6dB**), ...

They are described as examples (e.g.), but I think they should be required. Otherwise the decoder does not know the threshold values because they are not sent in the bitstream.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/608.html" title="Last updated on Jul 10, 2023, 11:26 PM UTC (e7801d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/608/75175af...e7801d6.html" title="Last updated on Jul 10, 2023, 11:26 PM UTC (e7801d6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2791,7 +2791,7 @@ Down-mix paths, which conform to the above rule, SHALL be only allowed for scala
 
 ### Annex B-4: Recon Gain Generation ### {#iamfgeneration-scalablechannelaudio-recongaingeneration}
 
-This section describes how to generate [=recon_gain=]. 
+This section RECOMMENDs how to generate [=recon_gain=]. 
 
 NOTE: Recon gain generation is not required when the codec is lossless such as [=codec_id=] is set to 'ipcm' or 'fLaC'.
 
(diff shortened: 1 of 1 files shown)


tdaede:
This is OK but it's still a bit unclear in the text whether the whole section is informative or just these values (and using RECOMMENDED just for those values makes it seem like the rest of the text isn't informative)

tdaede (approved):

PR #611, opened 2023-07-10
Apply restrictions to FLAC `decoder_config`.

jwcullen:
Apply restrictions to FLAC `decoder_config` similar to Opus `decoder_config`.

  - The `decoder_config` is shared between substreams. Some values do not make sense to apply to all substreams.
  
---------

Opus has these types of restrictions too. For example if we have a 5.1 FLAC audio element there will be an L/R stream with 2 channels and an LFE channel with 1 channel. They will both share the same `decoder_config`. 

The `minimum/maximum_block_size` restriction makes sure FLAC frames contain the same number of samples. The `minimum/maximum_frame_size` recommendation is because these values affect multiple sub streams that may be encoded with a different number of bytes.

***
<a href="https://pr-preview.s3.amazonaws.com/jwcullen/iamf/pull/611.html" title="Last updated on Jul 13, 2023, 3:14 AM UTC (55288a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/611/75175af...jwcullen:55288a3.html" title="Last updated on Jul 13, 2023, 3:14 AM UTC (55288a3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1703,7 +1703,7 @@ The sample rate used for computing offsets SHALL be the rate indicated by the [=
 - [=maximum block size=] SHALL be set to [=num_samples_per_frame=].
 - [=minimum frame size=] SHOULD be set to 0.
 - [=maximum frame size=] SHOULD be set to 0.
-- [=number of channels=] SHALL be set to 2. [=number of channels=] can be ignored because the real value can be determined from the [=Audio Element OBU=].
+- [=number of channels=] SHALL be set to 1. [=number of channels=] can be ignored because the real value can be determined from the [=Audio Element OBU=].
 - [=MD5 signature=] SHOULD be set to 0.
 
 The format of [=audio_frame()=] is [=FRAME=] of [[!FLAC]], which is composed of [=FRAME_HEADER=], followed by [=SUBFRAME=](s) (one [=SUBFRAME=] per channel) and followed by [=FRAME_FOOTER=].
--- a/index.bs
+++ b/index.bs
@@ -1700,8 +1700,8 @@ The sample rate used for computing offsets SHALL be the rate indicated by the [=
 [=decoder_config()=] for FLAC are the [=METADATA_BLOCK=]s of [[!FLAC]]. The [=METADATA_BLOCK_STREAMINFO=] has the following constraints:
 - [=minimum block size=] block SHALL be set to [=num_samples_per_frame=].
 - [=maximum block size=] block SHALL be set to [=num_samples_per_frame=].
-- [=minimum frame size=] block SHALL be set to 0.
-- [=maximum frame size=] block SHALL be set to 0.
+- [=minimum frame size=] block SHOULD be set to 0.
+- [=maximum frame size=] block SHOULD be set to 0.
 - [=number of channels=] block SHALL be set to 2. [=number of channels=] can be ignored because the real value can be determined from the [=Audio Element OBU=].
 
 
--- a/index.bs
+++ b/index.bs
@@ -127,6 +127,10 @@ url: https://xiph.org/flac/format.html; spec: FLAC; type: dfn;
 	text: maximum frame size
 	text: number of channels
 	text: MD5 signature
+	text: Block size in inter-channel samples
+	text: Sample rate
+	text: Channel assignment
+	text: Sample size in bits
 
 url: https://xiph.org/flac/format.html; spec: FLAC; type: property;
 	text: fLaC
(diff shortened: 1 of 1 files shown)


felicialim (approved):

jwcullen:
Hmm the MD5 signature field is also based on the audio streams. So it would be different per stream. Perhaps the spec should have a MUST or SHOULD be 0 here as well.

tdaede:
We will double check that this is easily addressable in the reference software.

I will check to make sure this is the correct method of indicating the lack of a MD5 in FLAC.

index.bs
```diff
@@ -1691,7 +1698,13 @@ The sample rate used for computing offsets SHALL be the rate indicated by the [=
 
 [=codec_id=] SHALL be 'fLaC', the FLAC stream marker in ASCII, meaning byte 0 of the stream is 0x66, followed by 0x4C 0x61 0x43.
 
-[=decoder_config()=] for FLAC is [=METADATA_BLOCK=] of [[!FLAC]].
+[=decoder_config()=] for FLAC are the [=METADATA_BLOCK=]s of [[!FLAC]]. The [=METADATA_BLOCK_STREAMINFO=] has the following constraints:
+- [=minimum block size=] SHALL be set to [=num_samples_per_frame=].
+- [=maximum block size=] SHALL be set to [=num_samples_per_frame=].
+- [=minimum frame size=] SHOULD be set to 0.
+- [=maximum frame size=] SHOULD be set to 0.
+- [=number of channels=] SHALL be set to 1. [=number of channels=] can be ignored because the real value can be determined from the [=Audio Element OBU=].
```
tdaede, line 1706: [=number of channels=] set to 1 means that this is a Stereo stream for FLAC (because FLAC defines this field as n+1)

tdaede:
We will clarify how this maps to SUBFRAMES (e.g. if we force the num_channels to 1 we will always have two SUBFRAMES).

index.bs
```diff
@@ -1691,7 +1698,13 @@ The sample rate used for computing offsets SHALL be the rate indicated by the [=
...
+- [=MD5 signature=] SHOULD be set to 0.
 
 The format of [=audio_frame()=] is [=FRAME=] of [[!FLAC]], which is composed of [=FRAME_HEADER=], followed by [=SUBFRAME=](s) (one [=SUBFRAME=] per channel) and followed by [=FRAME_FOOTER=].
```
sunghee-hwang, line 1709:
```suggestion
The format of [=audio_frame()=] is [=FRAME=] of [[!FLAC]] with the following constraints.
- [=Block size in inter-channel samples=] in the [FRAME_HEADER=] SHALL be set to [=num_samples_per_frame=].
- [=Sample rate=] in the [=FRAME_HEADER=] SHALL indicate the same sample rate defined in the [=METADATA_BLOCK_STREAMINFO=].
- [=Channel assignment=] in the [=FRAME_HEADER=] SHALL be set to 0 or 1 to indicate that the [=FRAME=] contains 1 channel or 2 channels, respectively.
- [=Sample size in bits=] in the [=FRAME_HEADER=] SHALL indicate the same sample size defined in the [=METADATA_BLOCK_STREAMINFO=]. 
```

index.bs
```diff
@@ -116,10 +116,17 @@ url: https://en.wikipedia.org/wiki/Q_(number_format); spec: Q-Format; type: dfn;
...
 	text: FRAME
 	text: FRAME_HEADER
 	text: SUBFRAME
 	text: FRAME_FOOTER
+	text: minimum block size
+	text: maximum block size
+	text: minimum frame size
+	text: maximum frame size
+	text: number of channels
+	text: MD5 signature
```
sunghee-hwang, line 129:
```suggestion
	text: MD5 signature
	text: Block size in inter-channel samples
	text: Sample rate
	text: Channel assignment
	text: Sample size in bits
```

sunghee-hwang (commented):
To clarify FLAC decoder_config and audio_frame()

index.bs
yilun-zhangs, line 1706: We have checked our IAMF encoder and decoder, I find our ref code has used the same way.
At decoder side, we will set the flac channels according to that it is coupled substream or single substream. 

So in my opinion, I agree to these modification, and it dose not affect our current IAMF decoder.

Thanks.

sunghee-hwang (approved):

tdaede:
I checked https://datatracker.ietf.org/doc/draft-ietf-cellar-flac/ and in section 8.2, indeed 0 is the correct "unknown md5" value.

PR #628, opened 2023-07-12
roundup() -> ceil(), for consistency with the existing floor()

felicialim:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/628.html" title="Last updated on Jul 12, 2023, 6:18 PM UTC (6ea66ea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/628/f047a0f...6ea66ea.html" title="Last updated on Jul 12, 2023, 6:18 PM UTC (6ea66ea)">Diff</a>

PR #627, opened 2023-07-12
Fix division in audio_roll_distance

felicialim:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/627.html" title="Last updated on Jul 12, 2023, 6:10 PM UTC (cc9b30b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/627/f047a0f...cc9b30b.html" title="Last updated on Jul 12, 2023, 6:10 PM UTC (cc9b30b)">Diff</a>

PR #630, opened 2023-07-12
Simplify division symbols

felicialim:
Drop "integer division with truncation of the result toward zero" since we can represent that with floor(x / y).

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/630.html" title="Last updated on Jul 17, 2023, 6:02 PM UTC (9701c04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/630/f047a0f...9701c04.html" title="Last updated on Jul 17, 2023, 6:02 PM UTC (9701c04)">Diff</a>


felicialim:
There are a few more uses of `/` in the sections below, but with this change, I think they're correct now (i.e. should be floating point division, not int divisions). @sunghee-hwang Could you please check as well?

7.2.1 Gain
7.2.2 De-Mixer
7.2.3 Recon Gain
10.2.4. Annex B-4: Recon Gain Generation

sunghee-hwang:
@felicialim, how about making PR #627, #628, #630, and #631 as one single PR or fully separated ones?

From past experience, we had lost something during the merging of multiple PRs complicating each other so we should create another PR to fix it.

sunghee-hwang:
> There are a few more uses of `/` in the sections below, but with this change, I think they're correct now (i.e. should be floating point division, not int divisions). @sunghee-hwang Could you please check as well?
> 
> 7.2.1 Gain 7.2.2 De-Mixer 7.2.3 Recon Gain 10.2.4. Annex B-4: Recon Gain Generation

Sure, I will.

sunghee-hwang:
> > There are a few more uses of `/` in the sections below, but with this change, I think they're correct now (i.e. should be floating point division, not int divisions). @sunghee-hwang Could you please check as well?
> > 7.2.1 Gain 7.2.2 De-Mixer 7.2.3 Recon Gain 10.2.4. Annex B-4: Recon Gain Generation
> 
> Sure, I will.

I have checked the sections, and they are fine. But, generally, I realize that there are many places to use '/' in other usages.
So, it would be better to use '÷' rather than '/'.

felicialim:
> @felicialim, how about making PR #627, #628, #630, and #631 as one single PR or fully separated ones?
> 
> From past experience, we had lost something during the merging of multiple PRs complicating each other so we should create another PR to fix it.

Alternatively, how about if we merge these smaller approved PRs offline before the weekly meeting? @tdaede

felicialim:
> > > There are a few more uses of `/` in the sections below, but with this change, I think they're correct now (i.e. should be floating point division, not int divisions). @sunghee-hwang Could you please check as well?
> > > 7.2.1 Gain 7.2.2 De-Mixer 7.2.3 Recon Gain 10.2.4. Annex B-4: Recon Gain Generation
> > 
> > 
> > Sure, I will.
> 
> I have checked the sections, and they are fine. But, generally, I realize that there are many places to use '/' in other usages. So, it would be better to use '÷' rather than '/'.

Do you mean we should still keep one division symbol only, but use "÷" instead of "/"? 

My preference is for "/" because it seems more natural to use "/" (from the high number of occurrence here, and from the mistakes we'd also caught in previous PRs) so this may help prevent accidental mistakes when making future extensions to the spec.

sunghee-hwang:
> > @felicialim, how about making PR #627, #628, #630, and #631 as one single PR or fully separated ones?
> > From past experience, we had lost something during the merging of multiple PRs complicating each other so we should create another PR to fix it.
> 
> Alternatively, how about if we merge these smaller PRs offline (after approval) before the weekly meeting? @tdaede

@felicialim @tdaede , I am fine if it would work.

sunghee-hwang:
> > > > There are a few more uses of `/` in the sections below, but with this change, I think they're correct now (i.e. should be floating point division, not int divisions). @sunghee-hwang Could you please check as well?
> > > > 7.2.1 Gain 7.2.2 De-Mixer 7.2.3 Recon Gain 10.2.4. Annex B-4: Recon Gain Generation
> > > 
> > > 
> > > Sure, I will.
> > 
> > 
> > I have checked the sections, and they are fine. But, generally, I realize that there are many places to use '/' in other usages. So, it would be better to use '÷' rather than '/'.
> 
> Do you mean we should still keep one division symbol only, but use "÷" instead of "/"?
> 
> My preference is for "/" because it seems more natural to use "/" (from the high number of occurrence here, and from the mistakes we'd also caught in previous PRs) so this may help prevent accidental mistakes when making future extensions to the spec.

Yes, if you search the spec by using the keyword '/', there are so many places to use it in different usage.

sunghee-hwang:
> > > > > There are a few more uses of `/` in the sections below, but with this change, I think they're correct now (i.e. should be floating point division, not int divisions). @sunghee-hwang Could you please check as well?
> > > > > 7.2.1 Gain 7.2.2 De-Mixer 7.2.3 Recon Gain 10.2.4. Annex B-4: Recon Gain Generation
> > > > 
> > > > 
> > > > Sure, I will.
> > > 
> > > 
> > > I have checked the sections, and they are fine. But, generally, I realize that there are many places to use '/' in other usages. So, it would be better to use '÷' rather than '/'.
> > 
> > 
> > Do you mean we should still keep one division symbol only, but use "÷" instead of "/"?
> > My preference is for "/" because it seems more natural to use "/" (from the high number of occurrence here, and from the mistakes we'd also caught in previous PRs) so this may help prevent accidental mistakes when making future extensions to the spec.
> 
> Yes, if you search the spec by using the keyword '/', there are so many places to use it in different usage.

@felicialim, did you contact him? How about texting or calling to him?

sunghee-hwang:
> > > > > There are a few more uses of `/` in the sections below, but with this change, I think they're correct now (i.e. should be floating point division, not int divisions). @sunghee-hwang Could you please check as well?
> > > > > 7.2.1 Gain 7.2.2 De-Mixer 7.2.3 Recon Gain 10.2.4. Annex B-4: Recon Gain Generation
> > > > 
> > > > 
> > > > Sure, I will.
> > > 
> > > 
> > > I have checked the sections, and they are fine. But, generally, I realize that there are many places to use '/' in other usages. So, it would be better to use '÷' rather than '/'.
> > 
> > 
> > Do you mean we should still keep one division symbol only, but use "÷" instead of "/"?
> > My preference is for "/" because it seems more natural to use "/" (from the high number of occurrence here, and from the mistakes we'd also caught in previous PRs) so this may help prevent accidental mistakes when making future extensions to the spec.
> 
> Yes, if you search the spec by using the keyword '/', there are so many places to use it in different usage.


@felicialim 
Let's go ahead with '/' as the industry has used it for multiple purposes. I will update the places which are using '/' in different usages if needed. Please reply to this when you read this.

felicialim:
> > > > > > There are a few more uses of `/` in the sections below, but with this change, I think they're correct now (i.e. should be floating point division, not int divisions). @sunghee-hwang Could you please check as well?
> > > > > > 7.2.1 Gain 7.2.2 De-Mixer 7.2.3 Recon Gain 10.2.4. Annex B-4: Recon Gain Generation
> > > > > 
> > > > > 
> > > > > Sure, I will.
> > > > 
> > > > 
> > > > I have checked the sections, and they are fine. But, generally, I realize that there are many places to use '/' in other usages. So, it would be better to use '÷' rather than '/'.
> > > 
> > > 
> > > Do you mean we should still keep one division symbol only, but use "÷" instead of "/"?
> > > My preference is for "/" because it seems more natural to use "/" (from the high number of occurrence here, and from the mistakes we'd also caught in previous PRs) so this may help prevent accidental mistakes when making future extensions to the spec.
> > 
> > 
> > Yes, if you search the spec by using the keyword '/', there are so many places to use it in different usage.
> 
> @felicialim Let's go ahead with '/' as the industry has used it for multiple purposes. I will update the places which are using '/' in different usages if needed. Please reply to this when you read this.

Sounds good. Most other "/" are hopefully obvious they aren't divisions, but I agree there could be a few places where it's ambiguous. Thanks for updating any that you see.

index.bs
```diff
@@ -579,7 +579,7 @@ NOTE: 'ipcm' should not be confused with 'lpcm' which is another 4CC to identify
...
 
 <dfn noexport>audio_roll_distance</dfn> indicates how many audio frames prior to the current audio frame need to be decoded (and decoded samples discarded) to set the encoder in a state that will produce the perfect decoded audio signal. It SHALL always be a negative value or zero. For some audio codecs, even if an audio frame can be decoded independently, the decoded signal after decoding only that frame may not represent a perfect, decoded audio signal, even ignoring compression artifacts. This can be due to overlap transforms. While potentially acceptable when starting to decode an [=Audio Substream=], it may be problematic when automatically switching between similar [=Audio Substream=]s of different quality and/or bitrate. 
-- It SHALL be set to -R when [=codec_id=] is set to 'Opus'. Where R is roundup(3840 % [=num_samples_per_frame=]).
+- It SHALL be set to -R when [=codec_id=] is set to 'Opus'. Where R is ceil(3840 ÷ [=num_samples_per_frame=]).
```
sunghee-hwang, line 582:
```suggestion
- It SHALL be set to -R when [=codec_id=] is set to 'Opus'. Where R is ceil(3840 / [=num_samples_per_frame=]).
```

sunghee-hwang (commented):
÷ to /

index.bs
```diff
@@ -2252,9 +2252,9 @@ where <code>B_quad(a) = (t, y)</code> is a 2D coordinate with parameter value <c
 To apply the parameter values to the audio samples in the subblock without interpolation, the [=parameter_rate=] SHOULD be first resampled to the audio sample rate to give:
 
 ```
-n0 = t0 * audio_sample_rate / parameter_rate,
-n1 = t1 * audio_sample_rate / parameter_rate,
-n2 = t2 * audio_sample_rate / parameter_rate.
+n0 = t0 * floor(audio_sample_rate / parameter_rate),
+n1 = t1 * floor(audio_sample_rate / parameter_rate),
+n2 = t2 * floor(audio_sample_rate / parameter_rate).
```
sunghee-hwang, line 2257:
```suggestion
n0 = floor(t0 * (audio_sample_rate / parameter_rate)),
n1 = floor(t1 * (audio_sample_rate / parameter_rate)),
n2 = floor(t2 * (audio_sample_rate / parameter_rate).
```

sunghee-hwang (changes requested):
Should it be like this?

sunghee-hwang (approved):

index.bs
```diff
@@ -2252,9 +2252,9 @@ where <code>B_quad(a) = (t, y)</code> is a 2D coordinate with parameter value <c
 To apply the parameter values to the audio samples in the subblock without interpolation, the [=parameter_rate=] SHOULD be first resampled to the audio sample rate to give:
 
 ```
-n0 = t0 * audio_sample_rate / parameter_rate,
-n1 = t1 * audio_sample_rate / parameter_rate,
-n2 = t2 * audio_sample_rate / parameter_rate.
+n0 = floor(t0 * (audio_sample_rate / parameter_rate)),
+n1 = floor(t1 * (audio_sample_rate / parameter_rate)),
+n2 = floor(t2 * (audio_sample_rate / parameter_rate).
```
sunghee-hwang, line 2257:
```suggestion
n2 = floor(t2 * (audio_sample_rate / parameter_rate)).
```

index.bs
```diff
@@ -579,7 +579,7 @@ NOTE: 'ipcm' should not be confused with 'lpcm' which is another 4CC to identify
...
 
 <dfn noexport>audio_roll_distance</dfn> indicates how many audio frames prior to the current audio frame need to be decoded (and decoded samples discarded) to set the encoder in a state that will produce the perfect decoded audio signal. It SHALL always be a negative value or zero. For some audio codecs, even if an audio frame can be decoded independently, the decoded signal after decoding only that frame may not represent a perfect, decoded audio signal, even ignoring compression artifacts. This can be due to overlap transforms. While potentially acceptable when starting to decode an [=Audio Substream=], it may be problematic when automatically switching between similar [=Audio Substream=]s of different quality and/or bitrate. 
-- It SHALL be set to -R when [=codec_id=] is set to 'Opus'. Where R is roundup(3840 % [=num_samples_per_frame=]).
+- It SHALL be set to -R when [=codec_id=] is set to 'Opus'. Where R is ceil(3840 / [=num_samples_per_frame=]).
```
sunghee-hwang, line 582:
```suggestion
- It SHALL be set to -R when [=codec_id=] is set to 'Opus', where R is <code>ceil(3840 / [=num_samples_per_frame=])</code>.
```

index.bs
```diff
@@ -579,7 +579,7 @@ NOTE: 'ipcm' should not be confused with 'lpcm' which is another 4CC to identify
 <dfn noexport>num_samples_per_frame</dfn> indicates the frame length, in samples, of the audio_frame() provided in by audio_frame_obu(). It SHALL NOT be set to zero. If the [=decoder_config()=] structure for a given codec specifies a value for the frame length, the two values SHALL be equal.
 
 <dfn noexport>audio_roll_distance</dfn> indicates how many audio frames prior to the current audio frame need to be decoded (and decoded samples discarded) to set the encoder in a state that will produce the perfect decoded audio signal. It SHALL always be a negative value or zero. For some audio codecs, even if an audio frame can be decoded independently, the decoded signal after decoding only that frame may not represent a perfect, decoded audio signal, even ignoring compression artifacts. This can be due to overlap transforms. While potentially acceptable when starting to decode an [=Audio Substream=], it may be problematic when automatically switching between similar [=Audio Substream=]s of different quality and/or bitrate.
```
sunghee-hwang, line 581:
```suggestion
<dfn noexport>audio_roll_distance</dfn> indicates how many audio frames prior to the current audio frame need to be decoded (and the decoded samples discarded) to set the encoder in a state that will produce the perfect decoded audio signal. It SHALL always be a negative value or zero. For some audio codecs, even if an audio frame can be decoded independently, the decoded signal after decoding only that frame may not represent a perfect, decoded audio signal, even ignoring compression artifacts. This can be due to overlap transforms. While potentially acceptable when starting to decode an [=Audio Substream=], it may be problematic when automatically switching between similar [=Audio Substream=]s of different quality and/or bitrate. 
```

index.bs
```diff
@@ -2012,7 +2012,7 @@ De-mixing for [=down-mixed audio=] for CL #i SHALL comply with the result by the
 - Top de-mixers
 	- <dfn noexport>TF2toT2 de-mixer</dfn>: Ltf2 = Ltf3 – w(k) x (L3 – L5) and Rtf2 = Rtf3 – w(k) x (R3 – R5)
 	- <dfn noexport>T2to4 de-mixer</dfn>: Ltb = 1/γ(k) x (Ltf2 – Ltf4) and Rtb = 1/γ(k) x (Rtf2 – Rtf4)
-- Where, Ltf2 / Rtf2 is top channel of x.1.2ch, Ltf3 / Rtf3 is top channel of 3.1.2ch, and Ltf4 / Rtf4 is top channel of x.1.4ch (x = 5 or 7) and [=w(k)=] is determined from the value of wIdx(k).
+- Where, Ltf2 and Rtf2 are top channels of x.1.2ch, Ltf3 and Rtf3 are top channels of 3.1.2ch, and Ltf4 and Rtf4 are top channels of x.1.4ch (x = 5 or 7) and [=w(k)=] is determined from the value of wIdx(k).
```
felicialim, line 2015:
```suggestion
- Where, Ltf2 and Rtf2 are the top channels of x.1.2ch, Ltf3 and Rtf3 are the top channels of 3.1.2ch, and Ltf4 and Rtf4 are the top channels of x.1.4ch (x = 5 or 7) and [=w(k)=] is determined from the value of wIdx(k).
```

Issue #687, opened 2023-08-03
§9.1.1: Datatypes definitions


podborski:
|Version|Source|Clause(s)|
|:---:|:---:|:---:|
|AOM Working Group Draft, 17 July 2023|Apple|§9.1.1|

#### Comment:
<blockquote>
leb128() syntaxName
</blockquote>

The way datatypes are defined seems problematic. Be consistent with other specifications of AOM (e.g. AV1)

<blockquote>
leb128()
</blockquote>

You should copy the definition form AV1 this one reads a bit strange. For example “the following unsigned integer”  is very strange because there is no “following”.

<blockquote>
string
</blockquote>

Missing a word after null-terminated?
Why don’t you use the definition form ISOBMFF?

-----
#### Suggestion:
Improve definitions and align with other specifications of AOM

PR #724, opened 2023-08-23
Fix #656, the maximum OBU size

### LINKED ISSUE #656: §3.2, §3.6: Max size of obu_header and audio_element_obu
|Version|Source|Clause(s)|
|:---:|:---:|:---:|
|AOM Working Group Draft, 17 July 2023|Apple|§3.2, §3.6|

#### Comment:
```
if (obu_extension_flag == 1) {
      leb128() extension_header_size;
      unsigned int (8*extension_header_size) extension_header_bytes;
    }
```
would allow 4GB extension header bytes. Any concerns here? Should it be restricted to a certain limit? Should it use fixed size length signaling which allows reasonable amount of size signaling?

Same question regarding audio_element_obu for param_definition_bytes and audio_element_config_bytes. which probably will not require max range of 4GB.

-----
#### Suggestion:
Clarify the range of OBU header and similar elements.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/724.html" title="Last updated on Aug 23, 2023, 5:57 AM UTC (89070fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/724/7faeb6f...89070fb.html" title="Last updated on Aug 23, 2023, 5:57 AM UTC (89070fb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1724,7 +1724,8 @@ IA decoders SHALL be able to parse all OBUs explicitly listed for this version o
 
 NOTE: In this section and subsections, the meaning of a unique OBU is that it is still unique if it only varies by the [=obu_redundant_copy=] flag.
 
-Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification:
+Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification: \(\left( 1 + n \right)^2\), for \(n = 0, 1, 2, \ldots, 14\)
+- The maximum size of an OBU (an obu_header followed by the OBU payload) SHALL be limited to \(2\text{MB}\) (i.e., \(2^21 \text{bytes}\)). It implies that the maximum value of the [=obu_size=] field SHALL be limited to \(2^21 - 4\). 
 - There SHALL be only one unique set of [=Descriptors=] in an [=IA Sequence=]. If the [=Descriptors=] are repeated in the middle of the [=IA Sequence=], all the OBUs in that set of [=Descriptors=] SHALL be marked as redundant (i.e., [=obu_redundant_copy=] = 1).
 	- When a set of [=Descriptors=] is placed in the middle of the [=IA Sequence=], it SHALL NOT be placed in the middle of a [=Temporal Unit=]. In other words, if [=Descriptors=] are placed mid-sequence, they SHALL be present only after the last OBU of a [=Temporal Unit=] and before the first OBU of the next [=Temporal Unit=].
 - There SHALL be only one unique [=Codec Config OBU=].
--- a/index.bs
+++ b/index.bs
@@ -1724,7 +1724,7 @@ IA decoders SHALL be able to parse all OBUs explicitly listed for this version o
 
 NOTE: In this section and subsections, the meaning of a unique OBU is that it is still unique if it only varies by the [=obu_redundant_copy=] flag.
 
-Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification: \(\left( 1 + n \right)^2\), for \(n = 0, 1, 2, \ldots, 14\)
+Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification:
 - The maximum size of an OBU (an obu_header followed by the OBU payload) SHALL be limited to \(2\text{MB}\) (i.e., \(2^21 \text{bytes}\)). It implies that the maximum value of the [=obu_size=] field SHALL be limited to \(2^21 - 4\). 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1724,7 +1724,8 @@ IA decoders SHALL be able to parse all OBUs explicitly listed for this version o
 
 NOTE: In this section and subsections, the meaning of a unique OBU is that it is still unique if it only varies by the [=obu_redundant_copy=] flag.
 
-Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification:
+Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification: \(\left( 1 + n \right)^2\), for \(n = 0, 1, 2, \ldots, 14\)
```
sunghee-hwang, line 1727:
```suggestion
Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification:
```

index.bs
```diff
@@ -1725,6 +1725,7 @@ IA decoders SHALL be able to parse all OBUs explicitly listed for this version o
 NOTE: In this section and subsections, the meaning of a unique OBU is that it is still unique if it only varies by the [=obu_redundant_copy=] flag.
 
 Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification:
+- The maximum size of an OBU (an obu_header followed by the OBU payload) SHALL be limited to \(2\text{MB}\) (i.e., \(2^21 \text{bytes}\)). It implies that the maximum value of the [=obu_size=] field SHALL be limited to \(2^21 - 4\).
```
sunghee-hwang, line 1728:
```suggestion
- The maximum size of an OBU (an obu_header followed by the OBU payload) SHALL be limited to \(2\text{MB}\) (i.e., \(2^{21} \text{bytes}\)). It implies that the maximum value of the [=obu_size=] field SHALL be limited to \(2^{21} - 4\). 
```

index.bs
```diff
@@ -1725,6 +1725,7 @@ IA decoders SHALL be able to parse all OBUs explicitly listed for this version o
 NOTE: In this section and subsections, the meaning of a unique OBU is that it is still unique if it only varies by the [=obu_redundant_copy=] flag.
 
 Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification:
+- The maximum size of an OBU (an obu_header followed by the OBU payload) SHALL be limited to \(2\text{MB}\) (i.e., \(2^{21} \text{bytes}\)). It implies that the maximum value of the [=obu_size=] field SHALL be limited to \(2^{21} - 4\).
```
sunghee-hwang, line 1728:
```suggestion
- The maximum size of an OBU (an obu_header followed by the OBU payload) SHALL be limited to \(2\text{MB}\) (i.e., \(2^{21}\) bytes). It implies that the maximum value of the [=obu_size=] field SHALL be limited to \(2^{21} - 4\). 
```

felicialim (approved):

index.bs
```diff
@@ -1725,6 +1725,7 @@ IA decoders SHALL be able to parse all OBUs explicitly listed for this version o
 NOTE: In this section and subsections, the meaning of a unique OBU is that it is still unique if it only varies by the [=obu_redundant_copy=] flag.
 
 Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification:
+- The maximum size of an OBU (an obu_header followed by the OBU payload) SHALL be limited to \(2\text{MB}\) (i.e., \(2^{21}\) bytes). It implies that the maximum value of the [=obu_size=] field SHALL be limited to \(2^{21} - 4\).
```
podborski, line 1728: Will it not be 2^{21} - 5? Looks like 1 byte of header before the `obu_size` was not taken into consideration?

sunghee-hwang: 4 was based on "1 byte (obu header before obu_size) + 3 bytes for obu_size field".
Did I miss anything?

jwcullen: Also consider the the possibility that `obu_size` is encoded to be larger than the minimum size with leading zeroes. 

In an old version we referenced AV1's definition of `leb128` which explicitly allowed this. Although the current IAMF spec is less clear here.

podborski: @sunghee-hwang you are right, its 3 bytes I mistakenly thought you will require 4 for the size signaling. The point on the leading zeros is worth clarification too. I also made this comment https://github.com/AOMediaCodec/iamf/issues/687 as the `leb128` definition was not clear to me either.

sunghee-hwang: Let me create another PR to address #687.

PR #748, opened 2023-08-31
Minor clarifications

### LINKED ISSUE #659: §3.2: On payload following the OBU size signalling
|Version|Source|Clause(s)|
|:---:|:---:|:---:|
|AOM Working Group Draft, 17 July 2023|Apple|§3.2|

#### Comment:
<blockquote>
obu_size indicates the size in bytes of the OBU immediately following the obu_size field of the OBU. An OBU MAY have extra bytes after consuming all the bytes per the OBU syntax definition. Parsers compliant with this version of the specification SHOULD ignore the extra bytes.
</blockquote>

Be careful, immediately following the obu_size will not be possible if obu_trimming_status_flag is set to 1 or if extension is present.

-----
#### Suggestion:
Align syntax with semantics

felicialim:
- obu_size includes the variable section of the OBU header (Fix #659)
- the extra bytes in the header to ignore refers to the extension bytes
- audio_roll_distance is related to obtaining the correct, not perfect,
  audio signal after decoding
- clarify that the leb128 size limit of 32 bits is on the value after
  leb128 parsing
- replace np.hanning reference with the explicit formula used by np.hanning

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -451,7 +451,7 @@ class OBUHeader() {
     leb128() num_samples_to_trim_at_end;
     leb128() num_samples_to_trim_at_start;
   }
-  if (obu_extension_flag == 1) {
+  if (obu_extension_flag) {
     leb128() extension_header_size;
     unsigned int (8 x extension_header_size) extension_header_bytes;
   }
@@ -506,7 +506,7 @@ This flag SHALL be set to 0 for this version of the specification. An OBU parser
 
 NOTE: A future version of the specification may use this flag to specify an extension header field by setting [=obu_extension_flag=] = 1 and setting the size of the extended header to [=extension_header_size=].
 
-<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU immediately following the obu_size field of the OBU. An OBU MAY have extra bytes after consuming all the bytes per the OBU syntax definition. Parsers compliant with this version of the specification SHOULD ignore the extra bytes.
+<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU immediately following the [=obu_size=] field. If the [=obu_trimming_status_flag=] and/or [=obu_extension_flag=] fields are set to 1, [=obu_size=] SHALL include the sizes of the additional fields.
 	
 <dfn noexport>num_samples_to_trim_at_end</dfn> indicates the number of samples that need to be trimmed from the end of the samples in this [=Audio Frame OBU=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -506,15 +506,15 @@ This flag SHALL be set to 0 for this version of the specification. An OBU parser
 
 NOTE: A future version of the specification may use this flag to specify an extension header field by setting [=obu_extension_flag=] = 1 and setting the size of the extended header to [=extension_header_size=].
 
-<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU immediately following the obu_size field of the OBU. An OBU MAY have extra bytes after consuming all the bytes per the OBU syntax definition. Parsers compliant with this version of the specification SHOULD ignore the extra bytes.
+<dfn noexport>obu_size</dfn> indicates the size in bytes of the OBU immediately following the [=obu_size=] field. If the [=obu_trimming_status_flag=] and/or [=obu_extension_flag=] fields are set to 1, [=obu_size=] SHALL include the sizes of the additional fields.
```
sunghee-hwang, line 509: @felicialim, don't you think that we still need these sentences?
"An OBU MAY have extra bytes after consuming all the bytes per the OBU syntax definition. Parsers compliant with this version of the specification SHOULD ignore the extra bytes."

felicialim: I believe these extra bytes are related to the extension header hence the edit below to extension_header_bytes. Right now, these sentences might be interpreted to mean extra bytes beyond obu_size, but I don't think this is intended, right?

sunghee-hwang: No. It is different from the header extension bytes. The extra bytes are something coming after the defined OBU syntax.
One example is that the future spec extends the OBU payload by adding some bytes at the end of the current OBU payload.

felicialim: Ok I see. Then how about the following modification:

obu_size indicates the size in bytes of the OBU immediately following the obu_size field. If the obu_trimming_status_flag and/or obu_extension_flag fields are set to 1, obu_size SHALL include the sizes of the additional fields. The obu_size MAY be greater than the size needed to represent the OBU syntax defined in this version of the specification, for example, to represent new syntax defined in a future version of the specification. Parsers compliant with this version of the specification SHOULD ignore the extra bytes.

sunghee-hwang: LGTM

sunghee-hwang (approved):

sunghee-hwang (approved):

PR #749, opened 2023-08-31
Fix #729, Reference for ref. s/w

### LINKED ISSUE #729: No reference to the reference software?
The reference software is part of the Deliverable but it doesn't seem to be mentioned in the document. There is also no explanation what the reference software is supposed to contribute or how the tests can be executed if that would be useful.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/749.html" title="Last updated on Aug 31, 2023, 11:09 PM UTC (ea3f080)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/749/c352f44...ea3f080.html" title="Last updated on Aug 31, 2023, 11:09 PM UTC (ea3f080)">Diff</a>


sunghee-hwang:
I don't know if this is the right way to do it. But it is my best.
Anyway, after getting confirmation for publication,
- The words, "working draft" needs to be removed
- The links for v1.0.0 and libiamf v1.0.0 need to be updated. (As there are no releases and no tags so far, they are just linking https://github.com/AOMediaCodec/iamf/ and https://github.com/AOMediaCodec/libiamf/, respectively.

felicialim:
LGTM for wording. I see that there is some redundancy with the "This version" and "Issue Tracking" metadata. I'll defer to the WG if the preference is to avoid this.

tdaede (approved):

tdaede:
This looks fine for text. I think there might be a better place in the document to place it where we get the box. I'm going to try to mess with it a bit and see if I can get it like the av1-spec doc.

tdaede:
Looking at the other StF WG specs also they have much simpler working, just linking to this and previous version, and also linking to the issue tracker independently in the metadata. We also have the Issue Tracking link already, so I think we can skip that text. I'll try to suggest a change.

index.bs
```diff
@@ -8,6 +8,8 @@ Repository: AOMediaCodec/iamf
 Shortname: iamf
 URL: https://aomediacodec.github.io/iamf/
 Date: 2023-08-30
+!Working Draft Version 1.0.0: This working draft version 1.0.0 of the IAMF Specification corresponds to the Git tag <a href="https://github.com/AOMediaCodec/iamf/">[v1.0.0]</a> in the <a href="https://github.com/AOMediaCodec/iamf/">[AOMediaCodec/IAMF]</a> project. Its content has been validated as consistent with the reference decoder provided by <a href="https://github.com/AOMediaCodec/libiamf/">[libiamf v1.0.0]</a>.
+!Working Draft Version 1.0.0: To comment on this document, use the <a href="https://github.com/AOMediaCodec/iamf/issues/">[issue tracker]</a>.
```
tdaede, line 12:
```suggestion
```

felicialim (approved):

tdaede:
Looks good.

Frank-Ba (approved):

PR #754, opened 2023-09-07
Fix #751, clarification on codecs and profiles

### LINKED ISSUE #751: On codecs
This review is based on the spec from Sept. 1st, 2023.

The way the specification is currently written regarding codecs has 2 aspects:
1. The `codecs` MIME parameter allows signaling of the underlying codec. So, content negotiation can be used to reject streams with unsupported codecs. Also, profiles are defined in a way that they are not immutable regarding codecs. If a next version of the spec adds a new codec, it will be acceptable with the current profiles. It seems that profiles and codecs are orthogonal. My suggestion is to make it clear. 
2. The current specification prevents derived specifications to use any other codecs than the 4 listed. This seems overly restrictive.

I would do the following:
1. add a NOTE at the beginning of section 4 saying: 
> NOTE: In this version of the specification, profiles impose constraints on how many codecs can be used in a bitstream but do not impose constraints on the actual codec used. In particular, this means that if a future version of the specification (or if a derived specification) defines how to use a new codec, the profiles defined in this specification could be used. Derived specifications may constrain the actual codec. The `codecs` parameter may also be used in content negotiation phases to ensure that a bitstream is supported by a device.

2. Rephrase as follows:
> For this version of the specification, it SHALL be set to one of the four [codec_id](https://aomediacodec.github.io/iamf/#codec_id) values defined below:

into
> This version of the specification defines support for the four [codec_id](https://aomediacodec.github.io/iamf/#codec_id) values defined below:
> NOTE: Derived specifications or future versions of this specification may define support for additional codecs.

3. And maybe in a future version of the specification, restructure the text so that codec-specific text is located in one place, maybe in a separate document pointed to by a registry, such as https://www.w3.org/TR/mse-byte-stream-format-registry/. This would make it easier for adoption of a new codec. The core spec would not have to be updated.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/754.html" title="Last updated on Sep 11, 2023, 10:35 PM UTC (5f910ad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/754/42194f0...5f910ad.html" title="Last updated on Sep 11, 2023, 10:35 PM UTC (5f910ad)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -596,7 +596,7 @@ class CodecConfig() {
 
 <dfn noexport>codec_config</dfn> is an instance of the [=CodecConfig()=] class, which provides codec-specific information for seting up the decoder.
 
-<dfn noexport>codec_id</dfn> indicates a ‘four-character code’ (4CC) to identify the codec used to generate the coded [=Audio Substream=]s. For this version of the specification, it SHALL be set to one of the four [=codec_id=] values defined below:
+<dfn noexport>codec_id</dfn> indicates a ‘four-character code’ (4CC) to identify the codec used to generate the coded [=Audio Substream=]s. This version of the specification defines support for the four [=codec_id=] values defined below:
 
 - <code>Opus</code>: All coded [=Audio Substream=]s referred to by all [=Audio Element=]s with this codec configuration SHALL comply with the [[!RFC-6716]] specification and the [=decoder_config=] structure SHALL comply with the constraints given in [[#opus-specific]].
 - <code>mp4a</code>: All coded [=Audio Substream=]s referred to by all [=Audio Element=]s with this codec configuration SHALL comply with the [[!AAC]] specification and the [=decoder_config=] structure SHALL comply with the constraints given in [[#aac-lc-specific]].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -605,6 +605,8 @@ class CodecConfig() {
 
 Parsers compliant with this version of the specification SHOULD ignore [=Codec Config OBU=]s with an unknown [=codec_id=].
 
+NOTE: Derived specifications or future versions of this specification may define support for additional codecs.
+
```
sunghee-hwang, line 609:
```suggestion
```

sunghee-hwang (commented):
Revert them because they are addressed by #756 to remove conflicts.

index.bs
```diff
@@ -1810,6 +1810,8 @@ The sample rate used for computing offsets SHALL be [=sample_rate=].
 
 # Profiles # {#profiles}
 
+NOTE: In this version of the specification, profiles impose constraints on how many codecs can be used in an [=IA Sequence=] but do not impose constraints on the actual codec used. In particular, this means that if a future version of the specification (or if a derived specification) defines how to use a new codec, the profiles defined in this specification could be used. Derived specifications may constrain the actual codec. The codecs parameter may also be used in content negotiation phases to ensure that an [=IA Sequence=] is supported by a device.
```
felicialim, line 1813: Does "codecs parameter" refer to Section 6.3? If so, it would be helpful to add a link (e.g. "[[#codecsparameter|codecs parameter]]").

felicialim (approved):

index.bs
sunghee-hwang, line 1813:
```suggestion
NOTE: In this version of the specification, profiles impose constraints on how many codecs can be used in an [=IA Sequence=] but do not impose constraints on the actual codec used. In particular, this means that if a future version of the specification (or if a derived specification) defines how to use a new codec, the profiles defined in this specification could be used. Derived specifications may constrain the actual codec. The [[#codecsparameter|codecs parameter]] may also be used in content negotiation phases to ensure that an [=IA Sequence=] is supported by a device.
```

cconcolato (approved):
The text is good. I would position the text after the first paragraph. It's a bit strange to start the section with a NOTE.

Issue #772, opened 2023-10-10
Generate PDF version of spec


tdaede:
We should generate a downloadable PDF version of the 1.0.0 spec now that it is approved.

sunghee-hwang:
There is no place to link IAMF spec in the current AOM home page (https://aomedia.org/av1-features/get-started/). It links for av1-related specs only.
I think that the AOM home page needs to be updated.

sunghee-hwang:
@tdaede @cconcolato 
As PR #775 has been merged, the link in IAMF PR indicates the working group draft.
https://aomedia.org/press%20releases/AOMedia-Advances-the-Audio-Innovation-Era/

The link ([Immersive Audio Model and Formats (IAMF)](https://aomediacodec.github.io/iamf/)) in IAMF PR needs to be updated to correctly indicate the final specification of IAMF v1.

cconcolato:
IAMF needs to adopt the proposed new headers (latest-approved, latest-draft, ...)

sunghee-hwang:
> IAMF needs to adopt the proposed new headers (latest-approved, latest-draft, ...)

Now, it's ready.

PR #775, opened 2023-11-07
convert to working draft as of Nov. 6

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/775.html" title="Last updated on Nov 7, 2023, 1:38 AM UTC (b8b8086)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/775/cd41a5c...b8b8086.html" title="Last updated on Nov 7, 2023, 1:38 AM UTC (b8b8086)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,14 +1,13 @@
 <pre class='metadata'>
 Group: AOM
-Status: FD
-Text Macro: SPECVERSION v1.0.0
+Status: WGD
 Title: Immersive Audio Model and Formats
 Editor: SungHee Hwang, Samsung, hshee@samsung.com
 Editor: Felicia Lim, Google, flim@google.com
 Repository: AOMediaCodec/iamf
 Shortname: iamf
 URL: https://aomediacodec.github.io/iamf/
-Date: 2023-10-03
+Date: 2023-11-06
 !Reference Implementation: <a href="https://github.com/AOMediaCodec/libiamf/releases/tag/v1.0.0/">libiamf v1.0.0</a>
 Abstract: This document specifies the Immersive Audio (IA) model, the standalone IA Sequence format, and the [[!ISO-BMFF]]-based IA container format.
 Local Boilerplate: footer yes


tdaede (approved):

sunghee-hwang:
@tdaede @cconcolato
As this PR has been merged, the link in IAMF PR indicates the working group draft.
https://aomedia.org/press%20releases/AOMedia-Advances-the-Audio-Innovation-Era/

The link in IAMF PR needs to be updated to correctly indicate the final specification of IAMF v1.
It may be related to #772.

felicialim:
> @tdaede @cconcolato As this PR has been merged, the link in IAMF PR indicates the working group draft. https://aomedia.org/press%20releases/AOMedia-Advances-the-Audio-Innovation-Era/
> 
> The link ([Immersive Audio Model and Formats (IAMF)](https://aomediacodec.github.io/iamf/)) in IAMF PR needs to be updated to correctly indicate the final specification of IAMF v1. It may be related to #772.

#777 and #778 should add 

1. A v1 html at https://aomediacodec.github.io/iamf/v1.0.0.html  
2. A "Previous Versions" field at the top of the doc that points to the v1.0.0 html

Would that work?

sunghee-hwang:
If we follow Cyril's suggestion (from the mail at the end of Oct.), they should look like this.

This version:
https://aomediacodec.github.io/iamf/index.html
Previously approved version:
https://aomediacodec.github.io/iamf/v1.0.0.html
Latest approved version:
https://aomediacodec.github.io/iamf/latest-approved.html
Latest draft version:
https://aomediacodec.github.io/iamf/latest-draft.html

However, I am not sure whether they are coded in index.bs file or not.
In case of av1-isobmff, they are not coded in index.bs file but index.html from aomediacodec.github.io/av1-isobmff shows them.

Unfortunately, I can not access "aomediacodec.github.io".

felicialim:
> If we follow Cyril's suggestion (from the mail at the end of Oct.), they should look like this.
Sorry I missed this!

> This version: https://aomediacodec.github.io/iamf/index.html Previously approved version: https://aomediacodec.github.io/iamf/v1.0.0.html Latest approved version: https://aomediacodec.github.io/iamf/latest-approved.html Latest draft version: https://aomediacodec.github.io/iamf/latest-draft.html
> 
> However, I am not sure whether they are coded in index.bs file or not. In case of av1-isobmff, they are not coded in index.bs file but index.html from aomediacodec.github.io/av1-isobmff shows them.

These look to have been manually added (https://github.com/AOMediaCodec/av1-isobmff/commit/ca087bd1391d9010faa7e5154e3697baa3ce1e9f). I've just done the same to v1.0.0.html to include "Latest draft version".

Since "Previous Versions" in index.bs can hold a list of multiple versions: an alternative is we add all approved versions there, so that it is auto-generated in future index.html files, and replaces "Previously/Latest approved version". If we prefer the av1-isobmff way, I can also make those manual changes.

sunghee-hwang:
Then I prefer the alternative from now on which means that we need to update the header manually in index.bs.
As we have already released v1, v1.0.0.html in gh-pages is manually updated as well.
Also, I prefer gh-pages having "latest-approved.html" and "lastest-draft.html" to keep aligned with av1-isobmff.

felicialim:
There is no latest-draft.html in the av1-isobmff main branch since there is no index.html to point to. This might be why v1.0.0.html in the main branch is different from the gh-pages branch. I think it would be better to keep both copies the same, if you agree, do you have a suggestion to work around this?

sunghee-hwang:
I agree that we need to keep both copies the same at least.

There are three places to store html files, main, gh-pages, and tags.
Even though we may not keep the same v1.0.0.html files, we need a structure to have the copies the same for future versions.
So, I would not mind if we could have such a structure.

I think that we may not need to care about the latest-draft.html in the main branch because the link (xxx/githu.io/iamf/xxx) seems relevant not to the main branch but to gh-pages.

What do you think?

PR #779, opened 2023-11-22
Fix #770, fix broken ambix reference

### LINKED ISSUE #770: Broken Ambix Reference
The link for https://aomediacodec.github.io/iamf/#biblio-ambix is broken:
https://iem.kug.ac.at/fileadmin/media/iem/projects/2011/ambisonics11_nachbar_zotter_sontacchi_deleflie.pdf

sunghee-hwang:


index.bs
```diff
@@ -208,9 +208,9 @@ url: https://www.iso.org/standard/77752.html#; spec: MP4-PCM; type: dfn;
 	},
 	"AmbiX": {
 		"title": "AMBIX - A SUGGESTED AMBISONICS FORMAT",
-		"status": "Paper",
+		"status": "Paper: Ambisonics Symposium, June 2011, Authors: Christian Nachbar, Franz Zotter, Etienne Deleflie, Alois Sontacchi",
```
felicialim, line 211: The SpecRefJsonFormat supports the "authors" key so we should probably use it. But the "publisher" key seems to be ignored in the generated index.html, so I think your hardcoding it as "Paper: Ambisonics Symposium, June 2011" is fine for a workaround.

sunghee-hwang: I tried but I failed. Can you make a commit suggestion?

sunghee-hwang:
It is closed as #790 replaces this PR.

PR #784, opened 2023-11-23
Fix #764, channel clarification

### LINKED ISSUE #764: Channel clarifications
The term 'channel' is never defined, but heavily used.

'Channel Group' is defined as a set of audio substreams. One could assume then that a channel is a substream? It would be wrong as for example, the definition coupled_substream_count talks about 'coupled stereo channels'. 

Because the audio_element_type values are `CHANNEL_BASED` and `SCENE_BASED`, one could assume that the term channel only applies to loudspeaker-based channels. This would also be wrong because for example the definition of ambisonics configuration talks about channels and profile restrictions on channel counts mixes ambisonics and others. 

So, the term channel seems overloaded, but I don't think we can change that. I would suggest defining it as:
> a component of a loudspeaker-based representation or a component of an Ambisonics representation.
> When used in the context of "channel group" or in the context of an audio element type (CHANNEL_BASED), it refers to loudspeaker-based channels. 

I would suggest updating Figure 2 to have 2 arrows between "Pre-Processor" and "Codec Encoder (and same between "Codec Decoder" and "Element Reconstructor": one arrow for Channel Groups, one arrow for Ambisonics Channels.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/784.html" title="Last updated on Nov 28, 2023, 1:36 AM UTC (002c7e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/784/cb5b4b2...002c7e2.html" title="Last updated on Nov 28, 2023, 1:36 AM UTC (002c7e2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -297,7 +297,9 @@ The model comprises a number of coded [=Audio Substream=]s and the metadata that
 
 The model includes one or more [=Audio Element=]s, each of which consists of one or more [=Audio Substream=]s. The [=Audio Substream=]s that make up an [=Audio Element=] are grouped into one or more [=Channel Group=]s. The model further includes [=Mix Presentation=]s and [=Parameter Substream=]s.
 
-The term <dfn noexport>3D audio signal</dfn> means a representation of sound that incorporates additional information beyond traditional stereo or surround sound formats such as Ambisonics (Scene-based), Object-based audio and Channel-based audio (e.g., 3.1.2ch or 7.1.4ch).
+The term <dfn noexport>3D audio signal</dfn> means a representation of sound that incorporates additional information beyond traditional stereo or surround sound formats such as Ambisonics (Scene-based audio), Object-based audio and Channel-based audio (e.g., 3.1.2ch or 7.1.4ch).
+
+NOTE: In this version of the specification, the term channel means a component of Scene-based audio, a component of Object-based audio, or a component of Channel-based audio. When used in the context of Channel-based audio, it refers to loudspeaker-based channels.
 
 The term <dfn noexport>Immersive Audio</dfn> (IA) means the combination of [=3D audio signal=]s recreating a sound experience close to that of a natural environment.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -297,7 +297,9 @@ The model comprises a number of coded [=Audio Substream=]s and the metadata that
 
 The model includes one or more [=Audio Element=]s, each of which consists of one or more [=Audio Substream=]s. The [=Audio Substream=]s that make up an [=Audio Element=] are grouped into one or more [=Channel Group=]s. The model further includes [=Mix Presentation=]s and [=Parameter Substream=]s.
 
-The term <dfn noexport>3D audio signal</dfn> means a representation of sound that incorporates additional information beyond traditional stereo or surround sound formats such as Ambisonics (Scene-based), Object-based audio and Channel-based audio (e.g., 3.1.2ch or 7.1.4ch).
+The term <dfn noexport>3D audio signal</dfn> means a representation of sound that incorporates additional information beyond traditional stereo or surround sound formats such as Ambisonics (Scene-based audio), Object-based audio and Channel-based audio (e.g., 3.1.2ch or 7.1.4ch).
+
+NOTE: In this version of the specification, the term channel means a component of Scene-based audio, a component of Object-based audio, or a component of Channel-based audio. When used in the context of Channel-based audio, it refers to loudspeaker-based channels.
```
cconcolato, line 302: Why not make it a defined term instead of a note?
Why use "in this version of the specification"

sunghee-hwang, line 302:
```suggestion
The term channel means a component of Scene-based audio, a component of Object-based audio, or a component of Channel-based audio. When used in the context of Channel-based audio, it refers to loudspeaker-based channels.
```

cconcolato (approved):

felicialim (approved):

PR #786, opened 2023-11-23
Fix #766, Duplicate values of audio_element_ids in one MixPresentatio…

### LINKED ISSUE #766: Duplicate values of audio_element_ids in the same loop should be prohibited
The profile restrictions say:
- in the common section:
> [num_audio_elements](https://aomediacodec.github.io/iamf/#num_audio_elements) SHOULD be set to 1 or 2

- in the simple profile section, the bitstream part:
> There SHALL be only one unique [Audio Element OBU](https://aomediacodec.github.io/iamf/#audio-element-obu).

- in the simple profile section, parser part:
> They SHALL be able to reconstruct one [Audio Element](https://aomediacodec.github.io/iamf/#audio-element).

The base profile section says:
> - They SHALL be able to reconstruct two [Audio Element](https://aomediacodec.github.io/iamf/#audio-element)s.
> - They SHALL be able to mix two [Audio Element](https://aomediacodec.github.io/iamf/#audio-element)s.

It talks about the "mix", which is missing in the first simple profile.
So I wonder if one could create a simple profile sequence with 1 audio element but with a submix that has num_audio_elements = 2, repeating the same value, which would not make sense.

I think the spec should say in the `num_audio_elements` semantics that a given value of `audio_element_id` shall not be repeated.

sunghee-hwang:
…n should be prohibited

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/786.html" title="Last updated on Nov 23, 2023, 12:28 PM UTC (818494b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/786/cb5b4b2...818494b.html" title="Last updated on Nov 23, 2023, 12:28 PM UTC (818494b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1176,7 +1176,7 @@ class MixPresentationOBU() {
 
 <dfn noexport>num_sub_mixes</dfn> specifies the number of sub-mixes. It SHALL NOT be set to 0. 
 
-<dfn noexport>num_audio_elements</dfn> specifies the number of [=Audio Element=]s that are used in this [=Mix Presentation=] to generate the final output audio signal for playback. It SHALL NOT be set to 0. 
+<dfn noexport>num_audio_elements</dfn> specifies the number of [=Audio Element=]s that are used in each sub-mix of this [=Mix Presentation=] to generate the final output audio signal for playback. It SHALL NOT be set to 0. There SHALL be no duplicate values of [=audio_element_obu/audio_element_id=] within one [=Mix Presentation=].
 
 <dfn noexport for="mix_presentation_obu">audio_element_id</dfn> indicates the identifier for an [=Audio Element=] which this [=Mix Presentation=] refers to.
 


index.bs
```diff
@@ -1176,7 +1176,7 @@ class MixPresentationOBU() {
 
 <dfn noexport>num_sub_mixes</dfn> specifies the number of sub-mixes. It SHALL NOT be set to 0. 
 
-<dfn noexport>num_audio_elements</dfn> specifies the number of [=Audio Element=]s that are used in this [=Mix Presentation=] to generate the final output audio signal for playback. It SHALL NOT be set to 0. 
+<dfn noexport>num_audio_elements</dfn> specifies the number of [=Audio Element=]s that are used in each sub-mix of this [=Mix Presentation=] to generate the final output audio signal for playback. It SHALL NOT be set to 0. There SHALL be no duplicate values of [=audio_element_obu/audio_element_id=] within one [=Mix Presentation=].
```
cconcolato, line 1179: Is it "within one [=Mix Presentation=]." or "within one sub-mix." ?

sunghee-hwang: It is "within one Mix Presentation".
Sub-mix concept has not been intended for such a purpose to share Audio Elemenets among sub-mixes.

felicialim (approved):

jwcullen:
Should this make it into the V1.0.0-Errata1? I think it should to make it clear this was not the intended behavior.

sunghee-hwang:
> Should this make it into the V1.0.0-Errata1? I think it should to make it clear this was not the intended behavior.

I agree with you.

PR #787, opened 2023-11-28
Fix #768, Improve scalable channel group and layer text

### LINKED ISSUE #768: Improve scalable channel group and layer text
The current text is spread:
- in the `audio_substream_id` semantics
- in 3.6.2. Scalable Channel Layout Config Syntax and Semantics, with some text mixed in the Syntax sub-section, before the Semantics sub-section
- in 3.6.2.1. Channel Layout Generation Rule and 3.6.2.2. Channel Group Format

It would be better if all of it was under a "scalable channel group and layout" section.

Also there are some typos in some formulas. For example:
"c = [1, ..., C] is the [Channel Group](https://aomediacodec.github.io/iamf/#channel-group) index and 
 is the number of [Channel Group](https://aomediacodec.github.io/iamf/#channel-group)s."
I think `=` should be `∈`. If `c` is an index, it cannot be equal to an array.
Same for `n_c`.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/787.html" title="Last updated on Nov 29, 2023, 1:38 AM UTC (badad83)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/787/cb5b4b2...badad83.html" title="Last updated on Nov 29, 2023, 1:38 AM UTC (badad83)">Diff</a>


index.bs
```diff
@@ -1029,35 +983,82 @@ Bit position : Channel Name
...
+\left[ 2, N_2 \right],
+\cdots,
+\left[ C, 1 \right],
+\left[ C, 2 \right],
+\cdots,
+\left[ C, N_c \right]
+\right]
+\]
+
+The order of the [=Audio Substream=]s in each [=Channel Group=] (i.e., the semantics of \(n_c\)) is specified in [[#syntax-scalable-channel-layout-config]].
```
felicialim, line 1061: I think this is referring to the paragraph starting "The order of the Audio Substream s in each Channel Group SHALL be as follows..."(?) 

It might be worth pulling that out into its own subsection too, and then referencing it directly here, and possibly in the definitions for `substream_count` and `coupled_substream_count`. (Could be a separate PR if you prefer)

felicialim (approved):

felicialim (approved):

PR #794, opened 2023-12-21
    Fix #792: Define a box for `configOBUs`.

### LINKED ISSUE #792: ISO-BMFF encapsulation should define a box inside the AudioSampleEntry for the configOBUs
Currently, the ISO-BMFF IAMF encapsulation is defined as follows:

```
class IASampleEntry extends AudioSampleEntry('iamf') {
    unsigned int (8) configOBUs[];
}
```

Said sample entry being in the stsd box. The problem here is that AudioSampleEntry can have other boxes inside it, like btrt (see section 8.5.2.2 of ISO 14496-12), and if said box is present, there's nothing delimiting where configOBUs ends. A parser may keep reading sample entry bytes after the last valid OBU, and will in turn try to parse a btrt box as if it was an OBU, with unpredictable results.

The solution is simple, albeit a problem considering the spec was finalized: Define a box inside the Sample Entry for the configOBUs, like it was done for example for AV1 in ISOBMFF (https://aomediacodec.github.io/av1-isobmff/#av1codecconfigurationbox-definition)

In short, something like:

```
class IASampleEntry extends AudioSampleEntry('iamf') {
    IAConfigurationBox config;
}

class IAConfigurationBox extends Box('iacb') {
    unsigned int (8) configOBUs[];
}
```

If changing how the existing Sample Entry is defined is not an option, then perhaps defining a new one and releasing a 1.0.1 version of the spec, stating it must be used when optional boxes within the AudioSampleEntry are present.

jwcullen:
- Future backwards compatible extensions can be put at the end of `IAConfigurationRecord`.
      - Future non-backwards compatible extensions will require `configurationVersion` to be bumped.

***
<a href="https://pr-preview.s3.amazonaws.com/jwcullen/iamf/pull/794.html" title="Last updated on Jan 18, 2024, 10:35 PM UTC (6f3007d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/794/b5668ae...jwcullen:6f3007d.html" title="Last updated on Jan 18, 2024, 10:35 PM UTC (6f3007d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -34,6 +34,7 @@ url: https://www.iso.org/standard/68960.html#; spec: ISO-BMFF; type: property;
 	text: trun
 	text: ctts
 	text: stss
+	text: btrt
 
 url: https://aomediacodec.github.io/av1-spec/av1-spec.pdf#; spec: AV1-Spec; type: dfn;
 	text: Clip3
--- a/index.bs
+++ b/index.bs
@@ -2016,7 +2016,7 @@ class IAConfigurationBox extends Box('iacb') {
 
 <b>Semantics</b>
 
-<dfn noexport>configurationVersion</dfn> indicates the version of the IAConfigurationBox. The value SHALL be set to 1. Parsers SHOULD ignore this box when [=configuraitonVersion=] is not set to 1.
+<dfn noexport>configurationVersion</dfn> indicates the version of the IAConfigurationBox. The value SHALL be set to 1 for this version of the specification. The box with which [=configurationVersion=] is not set to 1 SHALL be ignored by parsers compliant with this version of the specification.
 
 <dfn noexport>configOBUs_size</dfn> SHALL be set to the size of [=configOBUs=] in bytes.
 
@@ -2028,7 +2028,7 @@ class IAConfigurationBox extends Box('iacb') {
 
 NOTE: In practice, [=configOBUs=] is identical to [=Descriptors=].
 
-NOTE: `IAConfigurationBox` may contain extra data after the signaled end of [=configOBUs=]. Any extra data can be safely ignored.
+NOTE: Future versions of the specification may define fields after the signaled end of [=configOBUs=]. Parsers compliant with this version of the specification can safely ignore them.
 
 ### IA Sample Format ### {#iasampleformat}
 
--- a/index.bs
+++ b/index.bs
@@ -2026,7 +2026,7 @@ class IAConfigurationBox extends Box('iacb') {
 - One or more [=Audio Element OBU=]s
 - One or more [=Mix Presentation OBU=]s
 
-NOTE: In practice, [=configOBUs=] is the identical to [=Descriptors=].
+NOTE: In practice, [=configOBUs=] is identical to [=Descriptors=].
 
 NOTE: `IAConfigurationBox` may contain extra data after the signaled end of [=configOBUs=]. Any extra data can be safely ignored.
 
--- a/index.bs
+++ b/index.bs
@@ -1982,7 +1982,7 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
 
 ```
 class IASampleEntry extends AudioSampleEntry('iamf') {
-  unsigned int (8) configOBUs[];
+    IAConfigurationBox config;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jwcullen:
We probably want to allow some time to update the reference software and reference bitstream based on this change before we merge it in. But we can at least begin the initial review and work on any editorial fixes to the change in the meantime.

index.bs
```diff
@@ -1982,22 +1982,53 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
...
+  IAConfigurationRecord IAMFConfig;
+}
+```
+
+```
+aligned(8) class IAConfigurationRecord {
+    unsigned int (8) configurationVersion = 1;
+    leb128() configOBUs_size;
+    unsigned int (8) configOBUs[];
+}
```
sunghee-hwang, line 2015: I would suggest adding the syntaxes directly inside 'iacb', rather than inside IAConfigurationRecord, because IAConfigurationRecord itself is not extensible.

```suggestion
class IAConfigurationBox extends Box('iacb') {
    unsigned int (8) configurationVersion = 1;
    leb128() configOBUs_size;
    unsigned int (8) configOBUs[];
}
```

index.bs
```diff
@@ -1982,22 +1982,53 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
...
+```
+
 <b>Semantics</b>
 
+<dfn noexport>configurationVersion</dfn> indicates the version of the IAConfigurationRecord. The value SHALL be set to 1.
```
sunghee-hwang, line 2020:
```suggestion
<dfn noexport>configurationVersion</dfn> indicates the version of the IAConfigurationBox. The value SHALL be set to 1. Parsers SHOULD ignore this box when [=configuraitonVersion=] is not set to 1.
```

index.bs
```diff
@@ -1982,22 +1982,53 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
...
 Parsers SHALL ignore these two fields.
 
+### IA Configuration Box ### {#iaconfigurationbox-section}
+
+<pre class="def">
+	Sample Entry Type: <dfn export for="IAConfigurationBox">iacb</dfn>
+	Container:         Ia Sample Entry ('iamf')
+	Mandatory:         Yes
+	Quantity:          One or more.
+</pre>
```
sunghee-hwang, line 2000:
```suggestion
<pre class="def">
	Sample Entry Type: <dfn export for="IAConfigurationBox">iacb</dfn>
	Container:         IA Sample Entry ('iamf')
	Mandatory:         Yes
	Quantity:          One.
</pre>
```

index.bs
```diff
@@ -1982,22 +1982,53 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
...
+}
+```
+
 <b>Semantics</b>
 
+<dfn noexport>configurationVersion</dfn> indicates the version of the IAConfigurationRecord. The value SHALL be set to 1.
+
+<dfn noexport>configOBUs_size</dfn> SHALL be set to the size of [=configOBUs=] in bytes.
+
 <dfn noexport>configOBUs</dfn> SHALL contain the following OBUs in order.
```
sunghee-hwang, line 2024:
```suggestion
<dfn noexport>configOBUs</dfn> SHALL contain the following OBUs in order. It MAY contain Reserved OBUs after the [=IA Sequence Header OBU=] and before [=Mix Presentation OBU=]s.
```

index.bs
```diff
@@ -1982,22 +1982,53 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
...
 - [=IA Sequence Header OBU=]
 - [=Codec Config OBU=]
 - One or more [=Audio Element OBU=]s
 - One or more [=Mix Presentation OBU=]s
```
sunghee-hwang, line 2029:
```suggestion

NOTE: In practice, the [=configOBUs=] is the same as the [=Descriptors=], bit for bit.

```

index.bs
```diff
@@ -1982,22 +1982,53 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
...
 
+NOTE: `IAConfigurationRecord` may contain extra data after the signaled end of [=configOBUs=]. Any extra data can be safely ignored.
```
sunghee-hwang, line 2030:
```suggestion
NOTE: `IAConfigurationBox` may contain extra data after the signaled end of [=configOBUs=]. Any extra data can be safely ignored.
```

sunghee-hwang (commented):
No bitstream changes but editorial changes + more clarifications.

index.bs
jwcullen: Applied with slight rewording.

index.bs
```diff
@@ -1982,22 +1982,53 @@ NOTE: Multiple sample entries may be used in a track, for example when the track
...
 ```
 class IASampleEntry extends AudioSampleEntry('iamf') {
-  unsigned int (8) configOBUs[];
+    IAConfigurationBox config;
 }
 ```
 
 The [=channelcount=] field of [=AudioSampleEntry=] SHALL be set to 0. 
 The [=samplerate=] field of [=AudioSampleEntry=] SHALL be set to 0. There SHALL be no [=SamplingRateBox=].
 Parsers SHALL ignore these two fields.
```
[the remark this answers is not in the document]
jwcullen: Applied with slight rewording.

index.bs
jwcullen: I left this one out for now. I think to fully resolve the related issue this restriction may also need to be mentioned in 5.1.1, and we need to decide or clarify what to do with Temporal Delimiter OBUs.

Maybe better to do it in a different PR, since this one may get bloated with trying to do too much at once.

sunghee-hwang: Agreed. I will delete this suggestion.

sunghee-hwang (approved):

index.bs
```diff
@@ -1992,12 +1993,43 @@ Parsers SHALL ignore these two fields.
...
+class IAConfigurationBox extends Box('iacb') {
+    unsigned int (8) configurationVersion = 1;
+    leb128() configOBUs_size;
+    unsigned int (8) configOBUs[];
+}
+```
+
+<b>Semantics</b>
+
+<dfn noexport>configurationVersion</dfn> indicates the version of the IAConfigurationBox. The value SHALL be set to 1. Parsers SHOULD ignore this box when [=configuraitonVersion=] is not set to 1.
```
jamrial, line 2019: SHALL ignore IMO, because why would a parser even try to read what it doesn't understand? But that's nitpicking.

This should also mention the value of 1 is for streams complaint to this version of the spec. Similarly, the comment about parsers ignoring other values applies to those complaint to this version of the spec.

index.bs
```diff
@@ -1992,12 +1993,43 @@ Parsers SHALL ignore these two fields.
...
+
 <dfn noexport>configOBUs</dfn> SHALL contain the following OBUs in order.
 - [=IA Sequence Header OBU=]
 - [=Codec Config OBU=]
 - One or more [=Audio Element OBU=]s
 - One or more [=Mix Presentation OBU=]s
 
+NOTE: In practice, [=configOBUs=] is identical to [=Descriptors=].
+
+NOTE: `IAConfigurationBox` may contain extra data after the signaled end of [=configOBUs=]. Any extra data can be safely ignored.
```
jamrial, line 2031: Maybe something in line with other NOTEs in the document, like "Future versions of the specification MAY define fields after the signaled end of [=configOBUs=]. Parsers complaint to this version of the specification can be safely ignore them".

sunghee-hwang, line 2031:
```suggestion
NOTE: Future versions of the specification may define fields after the signaled end of [=configOBUs=]. Parsers compliant with this version of the specification can safely ignore them.
```

jwcullen:
I updated the notation based on feedback from the IAMF WG. It is just a style - not a bitstream change.

PR #804, opened 2024-04-05
Update spec header and add html file and images(folder) for the release of v1.0.0-errata

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/804.html" title="Last updated on Apr 11, 2024, 5:13 AM UTC (1869b1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/804/3c54ad0...1869b1f.html" title="Last updated on Apr 11, 2024, 5:13 AM UTC (1869b1f)">Diff</a>

PR #805, opened 2024-04-05
update symlink for the release of v1.0.0-errata

sunghee-hwang:


v1.0.0-errata.html
```diff
@@ -0,0 +1,7947 @@
...
+    <li data-md>
+     <p>Let the term "Audio Frames" mean the set of all <a data-link-type="dfn" href="#audio-frame-obu" id="ref-for-audio-frame-obu②④">Audio Frame OBU</a>s (for this <a data-link-type="dfn" href="#audio-element" id="ref-for-audio-element④④">Audio Element</a>) that have the same start timestamp. All Audio Frames in an <a data-link-type="dfn" href="#ia-sequence" id="ref-for-ia-sequence③⓪">IA Sequence</a> SHALL have the same number of <a data-link-type="dfn" href="#audio-frame-obu" id="ref-for-audio-frame-obu②⑤">Audio Frame OBU</a>s.</p>
+    <li data-md>
+     <p><a data-link-type="dfn" href="#parameter-block-obu" id="ref-for-parameter-block-obu①⑥">Parameter Block OBU</a>s MAY be associated with Audio Frames.</p>
+   </ul>
+   <center><img height="356" src="images/Immersive Audio Sequence with scalable channel audio (before OBU packing).png" style="width:100%; height:auto;" width="1622"></center>
```
cconcolato, line 2808: The links to images need to be fixed

cconcolato:
Can you make just one PR? with this one and #804.

sunghee-hwang:
> Can you make just one PR? with this one and #804.

I can add these files to #804 but I am not sure if those files will be reflected in this 'gh-pages' branch after merging them into the 'main' one.
When we merge something to the 'main' branch, it seems that the index.bs file is only reflected in this 'gh-pages' one.

v1.0.0-errata.html
sunghee-hwang: Done and now, all changes (including symlinks) for this PR are reflected in #804.

cconcolato:
We should not use gh-pages anymore. We should be using GitHub actions. See https://github.com/AOMediaCodec/av1-isobmff/blob/main/.github/workflows/auto-publish.yml for example

sunghee-hwang:
> We should not use gh-pages anymore. We should be using GitHub actions. See https://github.com/AOMediaCodec/av1-isobmff/blob/main/.github/workflows/auto-publish.yml for example

I replaced the auto-publish.yml in iamf with the one in av1-isobmff. (See #804)

tdaede (approved):

PR #824, opened 2024-05-28
Fix #817, leb128() is limited to 8 bytes.

### LINKED ISSUE #817: Clarify `leb128` padding limit
In a draft version of the spec we used normatively reference AV1 to define [`leb128`](https://aomediacodec.github.io/av1-spec/#leb128).

This was changed in https://github.com/AOMediaCodec/iamf/pull/558/files to bring in an explicit definition. Later a NOTE was added to clarify that "leading zero bytes" are allowed.

It seems like an oversight that we are missing a limit on the amount of zero-padded bytes.

> It is a requirement of bitstream conformance that the most significant bit of leb128_byte is equal to 0 if i is equal to 7. (This ensures that this syntax descriptor never uses more than 8 bytes.)

The benefit of allowing zero-padded bytes is to allow certain use-cases to leave `leb128`s blank, but fill them in arbitrarily later. But it seems prudent to limit the amount of zero-padded bytes. I think we intended this, but it was just missed in review.

I suggest adding language similar to the AV1 quote above to the IAMF definition of [`leb128()`](https://aomediacodec.github.io/iamf/#convention-data-types).

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/824.html" title="Last updated on May 28, 2024, 8:16 AM UTC (4f58927)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/824/ac94d8d...4f58927.html" title="Last updated on May 28, 2024, 8:16 AM UTC (4f58927)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2702,7 +2702,7 @@ All syntax elements conform to the [=Syntactic Description Language=] specified 
  
  <b>leb128()</b> indicates the type of an unsigned integer. To encode the following unsigned integer <b>syntaxName</b>, it first represents the integer in binary with an N-bit representation, where N is a multiple of 7. Then break the integer up into groups of 7 bits. Output one encoded byte for each 7 bits group, from least significant to most significant group. Each byte will have the group in its 7 least significant bits. Set the most significant bit on each byte except the last byte.
 
- <b>syntaxName</b> is an unsigned integer which is encoded by <b>leb128()</b>. The size of the unsigned integer to be encoded is limited to 32 bits. In other words, the value returned from the <b>leb128()</b> parsing process is less than or equal to \(2^{32} - 1\).
+ <b>syntaxName</b> is an unsigned integer which is encoded by <b>leb128()</b>. The size of the unsigned integer to be encoded is limited to 32 bits. In other words, the value returned from the <b>leb128()</b> parsing process is less than or equal to \(2^{32} - 1\). After encoding by <b>leb128()</b>, its size is limited to 8 bytes.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2702,7 +2702,7 @@ All syntax elements conform to the [=Syntactic Description Language=] specified
  
  <b>leb128()</b> indicates the type of an unsigned integer. To encode the following unsigned integer <b>syntaxName</b>, it first represents the integer in binary with an N-bit representation, where N is a multiple of 7. Then break the integer up into groups of 7 bits. Output one encoded byte for each 7 bits group, from least significant to most significant group. Each byte will have the group in its 7 least significant bits. Set the most significant bit on each byte except the last byte.
 
- <b>syntaxName</b> is an unsigned integer which is encoded by <b>leb128()</b>. The size of the unsigned integer to be encoded is limited to 32 bits. In other words, the value returned from the <b>leb128()</b> parsing process is less than or equal to \(2^{32} - 1\).
+ <b>syntaxName</b> is an unsigned integer which is encoded by <b>leb128()</b>. The size of the unsigned integer to be encoded is limited to 32 bits. In other words, the value returned from the <b>leb128()</b> parsing process is less than or equal to \(2^{32} - 1\). After encoding by <b>leb128()</b>, its size is limited to 8 bytes.
```
sunghee-hwang, line 2705:
```suggestion
 <b>syntaxName</b> is an unsigned integer which is encoded by <b>leb128()</b>. The size of the unsigned integer to be encoded is limited to 32 bits. In other words, the value returned from the <b>leb128()</b> parsing process is less than or equal to \(2^{32} - 1\). After encoding by <b>leb128()</b>, its maximum size is limited to 8 bytes.
```

felicialim (approved):

PR #826, opened 2024-05-28
Reword restriction on obu_size

### LINKED ISSUE #806: Reword restriction on `obu_size`
```
The maximum size of an OBU (an OBU Header followed by the OBU payload) SHALL be limited to 2MB (i.e., 
 2^21 bytes). It implies that the maximum value of the obu_size field SHALL be limited to 2^21 - 4.
```

 Clause 1) A maximum size of an entire OBU must be 2 MB or less.
 Clause 2) A maximum size of `obu_size` must be 2^21 - 4 or

When `obu_size` is written using the minimum number of bytes (i.e ` 0xfc, 0xff, 0x7f`). It takes up three bytes. The preceding portion of the header takes up one byte.
  - In this case an `obu_size` of 2^21 - 4 would correspond with an OBU which is 2MB.
  -  This clearly obeys clause 1 and clause 2.

When `obu_size` is written with padded bytes (e.g. `0xfc, 0xff, 0xff, 0x00` or  `0xfc, 0xff, 0xff, 0x80, 0x00` ). It may be more than three bytes.
  - In the first case an `obu_size` of 2^21 - 4 would correspond with an OBU that is 2 MB + 1 byte. Therefore we violate restriction 1. In order to obey both clause 1 and clause 2 `obu_size` would be limited to 2^21 - 5 (i.e. `0xfb, 0xff, 0xff, 0x00`). 
  - This is the max while obeying clause 1 and clause 2.

There is nothing that would violate clause 2 and NOT violate clause 1. I think the wording of clause 2 is confusing. But it does not need any technically changes.

felicialim:
Fix #806: Clarify the obu_size restriction holds if not encoded with padded
leb128 bytes

***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/826.html" title="Last updated on May 28, 2024, 6:25 PM UTC (a56796b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/826/ac94d8d...a56796b.html" title="Last updated on May 28, 2024, 6:25 PM UTC (a56796b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1827,7 +1827,7 @@ NOTE: [[#profiles-simple|Simple Profile]] or [[#profiles-base|Base Profile]] req
 NOTE: In this section and subsections, the meaning of a unique OBU is that it is still unique if it only varies by the [=obu_redundant_copy=] flag.
 
 Common restrictions on the [=IA Sequence=] for all profiles specified in this version of the specification:
-- The maximum size of an OBU (an [=OBU Header=] followed by the OBU payload) SHALL be limited to \(2\text{MB}\) (i.e., \(2^{21}\) bytes). It implies that the maximum value of the [=obu_size=] field SHALL be limited to \(2^{21} - 4\).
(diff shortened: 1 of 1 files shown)


jwcullen:
Thanks, this makes the note more precise.

sunghee-hwang (approved):
LGTM

PR #832, opened 2024-06-14
MixPresentationTags syntax and semantics

### LINKED ISSUE #815: Correct a typo on static 3.1.2ch down-mix matrix for 7.1.4ch of section 7.6.2
In the 3.1.2ch down-mix matrix for 7.1.4ch, the very right vector is currently 7.1.2ch but it should be converted to 7.1.4ch.

### LINKED ISSUE #816: Delete Recon_Gain_Flags.png (no more used) in images folder

### LINKED ISSUE #817: Clarify `leb128` padding limit

felicialim:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/832.html" title="Last updated on Jun 24, 2024, 8:27 PM UTC (b9cd8be)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/832/395610d...b9cd8be.html" title="Last updated on Jun 24, 2024, 8:27 PM UTC (b9cd8be)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1477,7 +1477,7 @@ class MixPresentationTags() {
   }
 
   leb128() mix_presentation_tags_extension_size;
-  unsigned int (8 * mix_presentation_tags_extension_size) mix_presentation_extension_bytes;
+  unsigned int (8 * mix_presentation_tags_extension_size) mix_presentation_tags_extension_bytes;
 }
 ```
 
--- a/index.bs
+++ b/index.bs
@@ -1505,9 +1505,9 @@ content_language    : Language of the audio content in this Mix Presentation.
 
 NOTE: The language indicated by [=tag_name=] = "content_language" is different from [=language_label=]. The former indicates the language of the audio content in the associated [=Audio Element=]s, while the latter indicates the language of the [=Mix Presentation=] annotations.
 
-<dfn noexport>mix_presentation_tags_extension_size</dfn> indicates the size in bytes of [=mix_presentation_extension_bytes=].
+<dfn noexport>mix_presentation_tags_extension_size</dfn> indicates the size in bytes of [=mix_presentation_tags_extension_bytes=].
 
-<dfn noexport>mix_presentation_extension_bytes</dfn> represents reserved bytes for future use. Parsers that don't understand these bytes SHOULD ignore them.
+<dfn noexport>mix_presentation_tags_extension_bytes</dfn> represents reserved bytes for future use. Parsers that don't understand these bytes SHOULD ignore them.
 
 
 ## Parameter Block OBU Syntax and Semantics ## {#obu-parameterblock}
--- a/index.bs
+++ b/index.bs
@@ -182,6 +182,12 @@ url: https://www.iso.org/standard/77752.html#; spec: MP4-PCM; type: dfn;
 		"publisher": "IETF",
 		"href": "https://www.rfc-editor.org/info/bcp47"
 	},
+	"ISO-639-2-Codes": {
+		"title": "ISO 639-2 Codes for the Representation of Names of Languages",
+		"status": "Standard",
+		"publisher": "ISO",
+		"href": "https://www.loc.gov/standards/iso639-2/php/code_list.php"
+	},
 	"FLAC": {
 		"title": "Free Lossless Audio Codec",
 		"status": "Best Practice",
@@ -1161,14 +1167,16 @@ class MixPresentationOBU() {
       ElementMixConfig element_mix_config;
     }
     OutputMixConfig output_mix_config;
-    
+
     leb128() num_layouts;
     for (j = 0; j < num_layouts; j++) {
       Layout loudness_layout;
       LoudnessInfo loudness; 
     }
   }
-}  
+
+  MixPresentationTags mix_presentation_tags;
+}
 ```
 
 <b>Semantics</b>
@@ -1214,6 +1222,9 @@ If a sub-mix in a [=Mix Presentation OBU=] includes only one single scalable cha
 
 	- The highest [=loudness_layout=] specified in one sub-mix is the layout that was used for authoring the sub-mix. The exception is when the [=Audio Element=] is a zero-order Ambisonics or Mono channel.
 	- The highest [=loudness_layout=] for a zero-order Ambisonics or Mono channel [=Audio Element=] is Stereo.
+
+<dfn noexport>mix_presentation_tags</dfn> is an instance of the [=MixPresentationTags()=] class, which provides informational metadata about a Mix Presentation, in addition to [=mix_presentation_annotations=].
+
  
 ### Mix Presentation Annotations Syntax and Semantics ### {#obu-mixpresentation-annotation}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1452,6 +1463,53 @@ NOTE: [[!ITU-1770-4]] adopts the convention of using the dBov unit for dBFS, whe
...
+<dfn noexport>tag_name</dfn> is the label describing a [=Mix Presentation=] tag. Parsers that don't understand a [=tag_name=] SHOULD ignore it and its corresponding [=tag_value=].
+
+This specification supports the following [=tag_name=]s:
+
+<pre class = "def">
+tag_name            : Description
+content_language    : Language of the audio content in this Mix Presentation.
+</pre>
+
+- There SHALL be at most one instance of [=tag_name=] = "content_language" within one [=Mix Presentation=]. If there are two or more instances, parsers SHOULD use the [=tag_value=] corresponding to the first instance, and MAY ignore the remaining instances.
```
sunghee-hwang, line 1497: "If there are two or more instances" What does it mean?
Does it mean that "If there are two or more instances of [=tag_name=] = "content_language""? or
"If there are multiple [=tag_name=]s"?

In the latter case, when a future spec defines a new tag_name, does it mean that "content_language" SHALL be followed by the new tag_name?

index.bs
```diff
@@ -1452,6 +1463,53 @@ NOTE: [[!ITU-1770-4]] adopts the convention of using the dBov unit for dBFS, whe
...
+
+<dfn noexport>tag_value</dfn> is the value of a [=Mix Presentation=] tag.
+
+- If the corresponding [=tag_name=] = "content_language", the following applies to this [=tag_value=].
+	- It indicates the language of the audio content in all [=Audio Elements=] within this [=Mix Presentation=].
+	- It SHALL conform to [[!BCP-47]].
+	- If a [=Mix Presentation=] contains [=Audio Element=]s with different language content, its corresponding [=tag_value=] SHOULD use one of the following [[!ISO-639-2-Codes]] language codes: <code>und</code> or <code>mul</code>.
```
sunghee-hwang, line 1504: I realize that ISO-639-2-Codes is better reference than BCP-47 in this case.
```suggestion
- If the corresponding [=tag_name=] = "content_language", the following applies to this [=tag_value=].
	- It indicates the language of the audio content in all [=Audio Element=]s within this [=Mix Presentation=].
	- It SHALL conform to [[!ISO-639-2-Codes]].
	- If a [=Mix Presentation=] contains [=Audio Element=]s with different language content, its corresponding [=tag_value=] SHOULD use one of the following [[!ISO-639-2-Codes]] language codes: <code>und</code> or <code>mul</code>.
```

index.bs
```diff
@@ -1452,6 +1463,53 @@ NOTE: [[!ITU-1770-4]] adopts the convention of using the dBov unit for dBFS, whe
...
+
+NOTE: The language indicated by [=tag_name=] = "content_language" is different from [=language_label=]. The former indicates the language of the audio content in the associated [=Audio Element=]s, while the latter indicates the language of the [=Mix Presentation=] annotations.
+
+<dfn noexport>mix_presentation_tags_extension_size</dfn> indicates the size in bytes of [=mix_presentation_extension_bytes=].
+
+<dfn noexport>mix_presentation_extension_bytes</dfn> represents reserved bytes for future use. Parsers that don't understand these bytes SHOULD ignore them.
```
sunghee-hwang, line 1510:
```suggestion
<dfn noexport>mix_presentation_tags_extension_size</dfn> indicates the size in bytes of [=mix_presentation_tags_extension_bytes=].

<dfn noexport>mix_presentation_tags_extension_bytes</dfn> represents reserved bytes for future use. Parsers that don't understand these bytes SHOULD ignore them.
```

index.bs
```diff
@@ -1452,6 +1463,53 @@ NOTE: [[!ITU-1770-4]] adopts the convention of using the dBov unit for dBFS, whe
...
+```
+class MixPresentationTags() {
+  unsigned int (8) num_name_value_tags;
+  for (int i = 0; i < num_name_value_tags; i++) {
+    string tag_name;
+    string tag_value;
+  }
+
+  leb128() mix_presentation_tags_extension_size;
+  unsigned int (8 * mix_presentation_tags_extension_size) mix_presentation_extension_bytes;
```
sunghee-hwang, line 1480: Are they necessarily required?

sunghee-hwang (commented):
Clarification

index.bs
```diff
@@ -1161,14 +1167,16 @@ class MixPresentationOBU() {
...
+
     leb128() num_layouts;
     for (j = 0; j < num_layouts; j++) {
       Layout loudness_layout;
       LoudnessInfo loudness; 
     }
   }
-}  
+
+  MixPresentationTags mix_presentation_tags;
```
sunghee-hwang, line 1178: Does every Mix Presentation include MixPresentationTags? or Does it allow Mix Presentation with no MixPresentationTags?

index.bs
sunghee-hwang, line 1480:
```suggestion
  unsigned int (8 * mix_presentation_tags_extension_size) mix_presentation_tags_extension_bytes;
```

index.bs
felicialim: The intention was that Mix Presentations should indicate only one content language. I'll edit this to clarify it's one instance of  [=tag_name=] = "content_language". 

There can be multiple [=tag_name=]s.

index.bs
felicialim: FYI, I had used "all [=Audio Element=]s" to emphasize that a Mix Presentation should only indicate one language. I'm ok with changing to "the associated [=Audio Element=]s" as long as you also don't think this distinction is lost. 

I'll commit this change for now, but please feel free to amend if anything changes based on the above.

index.bs
felicialim: No, they are not required. This would be indicated with `num_name_value_tags = 0`.

sunghee-hwang: Isn't it slightly different?
Mix Presentation with No MixPresentationTags vs. Mix Presentation with MixPresentationTags with num_name_value_tags = 0.
Please clarify it more.

felicialim: Ah sorry, I thought the question was about whether we conceptually allow a Mix Presentation with no tags.

In the bitstream, every Mix Presentation must have one instance of MixPresentationTags (which can have num_name_value_tags = 0). Do you have a preference otherwise?

index.bs
sunghee-hwang: There would be no problem with future updates such as we are currently adding MixPresentationTags as this is the last part of the OBU. I prefer not to have them.

index.bs
sunghee-hwang: I just wanted to know its intention.

sunghee-hwang: Let me share my thoughts about this.

Even though we mandate MixPresentationTags in v1.1, we have two types of Mix Presentation OBUs (MPO). One is an MPO of v1.0 and the other is an MPO with this class (let's call it class A). Only obu_size can indicate what the type of an MPO is.
So, we can't help allowing that MixPresenationTags is Optional.

Under this situation, if we need to consider another class() to be added in the future (let's call it class B), parsers can't distinguish an MPO with class A from an MPO with class B under this syntax structure.

When obu_size is greater than the size up to the end of num_sub_mixes loop, the first syntax right after the end of num_sub_mixes loop SHALL be recognizable. 

I think that we need to take one of two options below.

[Option1]
They can only distinguish an MPO vs an MPO with class A vs. an MPO with class A followed by class B.
For this, we need to clarify that the presence of MixPresentationTags is optional. If the obu_size is greater than this size up to the end of num_sub_mixes loop, it indicates that MixPresentationTags SHALL be present.
(The future spec having class B will specify that when class B is present, class A SHALL be followed by class B).

[Option2]
Another option is to define its structure as num_parameters but as an optional class. If the obu_size is greater than this size up to the end of num_sub_mixes loop, it indicates that MixPresentationExtension SHALL be present.
```
MixPresentationExtension() {
  leb128() num_extensions; // it is still allowed to be set to 0.
  for (i = 0; i < num_extensions; i++) {
    leb128() extension_type;
    if (extension_type == Mix_Presentation_Tags) {
      MixPresentationTags mix_presentation_tags;
    }
    else {
      leb128() extension_size;
      unsigned int (8 x extension_size) extension_bytes;
    }
  }
}
```

What do you think of this?

felicialim: Thanks for explanation and suggestions. Both options sound reasonable. With Option 2, if we do want to make a particular future extension_type mandatory, we could specify that via Profiles.

index.bs
felicialim: Sounds good, it's been dropped. Since the `MixPresentationTags` class will not be extensible, I think it should be also fine to shorten the naming of `num_name_value_tags` to `num_tags`.

index.bs
sunghee-hwang, line 1178:
```suggestion
  MixPresentationTags mix_presentation_tags; // Optional class
```

index.bs
```diff
@@ -1214,6 +1222,9 @@ If a sub-mix in a [=Mix Presentation OBU=] includes only one single scalable cha
 
 	- The highest [=loudness_layout=] specified in one sub-mix is the layout that was used for authoring the sub-mix. The exception is when the [=Audio Element=] is a zero-order Ambisonics or Mono channel.
 	- The highest [=loudness_layout=] for a zero-order Ambisonics or Mono channel [=Audio Element=] is Stereo.
+
+<dfn noexport>mix_presentation_tags</dfn> is an instance of the [=MixPresentationTags()=] class, which provides informational metadata about a Mix Presentation, in addition to [=mix_presentation_annotations=].
+
```
sunghee-hwang, line 1227:
```suggestion
<dfn noexport>mix_presentation_tags</dfn> is an instance of the [=MixPresentationTags()=] class, which provides informational metadata about a Mix Presentation, in addition to [=mix_presentation_annotations=].

The [=MixPresentationTags()=] class MAY or MAY NOT be present in a [=Mix Presentation OBU=]. If the [=obu_size=] of a [=Mix Presentation OBU=] is greater than the size up to the end of [=num_sub_mixes=] loop, the [=MixPresentationTags()=] SHALL be present in the [=Mix Presentation OBU=]. For a given [=IA Sequence=] with multiple [=Mix Presentation OBU=]s, the [=MixPresentationTags()=] MAY be present in some [=Mix Presentation OBU=]s and MAY NOT be present in the other [=Mix Presentation OBU=]s. 
```

sunghee-hwang (commented):
Clarification for MixPresentationTags

sunghee-hwang (approved):

PR #879, opened 2025-05-26
fix typo (leb128 to leb128())

sunghee-hwang:


sunghee-hwang:
It is replaced by PR #881

PR #884, opened 2025-06-12
fix #869: The reference for FLAC is changed into RFC-9639

### LINKED ISSUE #869: Update FLAC references to use RFC 9639
Looks like FLAC migrated to [RFC 9639](https://datatracker.ietf.org/doc/rfc9639/).

Tracking this here in case we want to update the future versions of the IAMF spec to refer to the RFC instead of the website.

sunghee-hwang:
***
<a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/pull/884.html" title="Last updated on Jun 12, 2025, 11:11 PM UTC (0299099)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/AOMediaCodec/iamf/884/5b35e5a...0299099.html" title="Last updated on Jun 12, 2025, 11:11 PM UTC (0299099)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -112,7 +112,7 @@ url: https://tools.ietf.org/html/rfc9639#; spec: RFC-9639; type: dfn;
 	text: maximum frame size
 	text: Sample rate
 	text: (number of channels) - 1
-	text: (bits per sample) - 1
+	text: bits per sample
 	text: MD5 checksum
 	text: Block Size Bits
 	text: Sample Rate Bits
--- a/index.bs
+++ b/index.bs
@@ -101,21 +101,23 @@ url: https://www.itu.int/rec/R-REC-BS.2051#; spec: ITU-2051-3; type: dfn;
 	text: Loudspeaker configuration for Sound System J (4+7+0)
 	text: SP Label
 
-url: https://xiph.org/flac/format.html; spec: FLAC; type: dfn;
-	text: METADATA_BLOCK
-	text: METADATA_BLOCK_STREAMINFO
-	text: FRAME
-	text: FRAME_HEADER
+url: https://tools.ietf.org/html/rfc9639#; spec: RFC-9639; type: dfn;
+	text: Metadata Block
+	text: Streaminfo
+	text: Frame
+	text: Frame Header
 	text: minimum block size
 	text: maximum block size
 	text: minimum frame size
 	text: maximum frame size
-	text: number of channels
-	text: MD5 signature
-	text: Block size in inter-channel samples
 	text: Sample rate
-	text: Channel assignment
-	text: Sample size in bits
+	text: (number of channels) - 1
+	text: (bits per sample) - 1
+	text: MD5 checksum
+	text: Block Size Bits
+	text: Sample Rate Bits
+	text: Channels Bits
+	text: Bit Depth Bits
 
 url: https://www.iso.org/standard/77752.html#; spec: MP4-PCM; type: dfn;
 	text: format_flags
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -101,21 +101,23 @@ url: https://www.itu.int/rec/R-REC-BS.2051#; spec: ITU-2051-3; type: dfn;
...
 	text: minimum frame size
 	text: maximum frame size
-	text: number of channels
-	text: MD5 signature
-	text: Block size in inter-channel samples
 	text: Sample rate
-	text: Channel assignment
-	text: Sample size in bits
+	text: (number of channels) - 1
+	text: (bits per sample) - 1
```
felicialim, line 115: I don't think this is used?

index.bs
```diff
@@ -1848,21 +1850,21 @@ The sample rate used for computing offsets SHALL be the rate indicated by the [=
...
 
-The [=DecoderConfig()=] class for FLAC is the [=METADATA_BLOCK=]s of [[!FLAC]] for mono or stereo channels. The [=METADATA_BLOCK_STREAMINFO=] has the following constraints:
+The [=DecoderConfig()=] class for FLAC is the [=Metadata Block=]s of [[!RFC-9639]] for mono or stereo channels. The [=Streaminfo=] has the following constraints:
 - [=minimum block size=] SHALL be set to [=num_samples_per_frame=].
 - [=maximum block size=] SHALL be set to [=num_samples_per_frame=].
 - [=minimum frame size=] SHOULD be set to 0.
 - [=maximum frame size=] SHOULD be set to 0.
-- [=number of channels=] SHALL be set to 1. [=number of channels=] can be ignored because the real value can be determined from the [=Audio Element OBU=] and from the [=Frame_Header=].
-- [=MD5 signature=] SHOULD be set to 0.
+- [=(number of channels) - 1=] SHALL be set to 1. [=(number of channels) - 1=] can be ignored because the real value can be determined from the [=Audio Element OBU=] and from the [=Frame Header=].
```
felicialim, line 1858: was it intentional to add "- 1"? Even though the value here is ignored, it changes what SHALL be used in the bitstream. Or maybe since it's not used, the requirement could be dropped.

index.bs
```diff
@@ -1848,21 +1850,21 @@ The sample rate used for computing offsets SHALL be the rate indicated by the [=
...
-- [=Block size in inter-channel samples=] in the [=FRAME_HEADER=] SHALL be set to [=num_samples_per_frame=].
-- [=Sample rate=] in the [=FRAME_HEADER=] SHALL indicate the same sample rate defined in the [=METADATA_BLOCK_STREAMINFO=].
-- [=Channel assignment=] in the [=FRAME_HEADER=] SHALL be set to 0 or 1 to indicate that the [=FRAME=] contains mono channel or stereo channels, respectively.
-- [=Sample size in bits=] in the [=FRAME_HEADER=] SHALL indicate the same sample size defined in the [=METADATA_BLOCK_STREAMINFO=]. 
+The format of [=audio_frame=] is [=Frame=] of [[!RFC-9639]] which contains only one single frame of mono or stereo channels with the following constraints.
+- [=Block Size Bits=] in the [=Frame Header=] SHALL be set to [=num_samples_per_frame=].
+- [=Sample Rate Bits=] in the [=Frame Header=] SHALL indicate the same sample rate defined in the [=Streaminfo=].
+- [=Channel Bits=] in the [=Frame Header=] SHALL be set to 0 or 1 to indicate that the [=FRAME=] contains mono channel or stereo channels, respectively.
+- [=Bit Depth Bits=] in the [=Frame Header=] SHALL indicate the same bits per sample defined in the [=Streaminfo=].
```
felicialim, line 1865: was this intended to be [=bits per sample=]?

index.bs
sunghee-hwang, line 115:
```suggestion
	text: bits per sample
```

index.bs
sunghee-hwang: Yes, it was intended. It follows the same syntax as RFC-9639. 
However, there will be no change in the bitstream because it will still be set to 1 to indicate stereo.

